PR #3, opened 2018-04-11
attempt a not_converted_yet test within framework

kerwinso:
@m1yag1: I know this code isn't working, but not sure why. What I'm trying to do is locate a `<div>` with `class='not converted yet'`. The test would be to make sure that a page does _not_ have this class in the content (i.e., this is _undesired_ content).

In the test, I tried jimmying in a URL I knew has this class in it (https://qa.cnx.org/contents/26b372ce-28e5-4338-95e6-49639cd2a88d). But when I run pytest on `test_cnx_content.py`, the test is passing. It's supposed to fail with that particular URL.

### DIFF
--- a/pages/content.py
+++ b/pages/content.py
@@ -10,7 +10,7 @@ from pages.base import Base
 class Content(Base):
     _content_locator = (By.ID, 'content')
     _title_locator = (By.CSS_SELECTOR, '.media-title h1')
-    _ncy_locator = (By.CSS_SELECTOR, 'not-converted-yet')
+    _ncy_locator = (By.CLASS_NAME, 'not-converted-yet')
 
     def wait_for_page_to_load(self):
         self.wait.until(lambda s: self.is_element_displayed(
@@ -20,10 +20,6 @@ class Content(Base):
     @property
     def title(self):
         return self.find_element(*self._title_locator).text
-
-    @property
-    def not_converted_yet(self):
-        return self.find_element(*self._ncy_locator).text
 
     @property
     def is_ncy_displayed(self):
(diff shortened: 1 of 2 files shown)


pages/content.py
```diff
@@ -10,6 +10,7 @@
 class Content(Base):
     _content_locator = (By.ID, 'content')
     _title_locator = (By.CSS_SELECTOR, '.media-title h1')
+    _ncy_locator = (By.CSS_SELECTOR, 'not-converted-yet')
```
m1yag1: If it's a class you need to add the `.` at the beginning. So this should be `.not-converted-yet`

pages/content.py
```diff
@@ -19,3 +20,11 @@ def wait_for_page_to_load(self):
     @property
     def title(self):
         return self.find_element(*self._title_locator).text
+
+    @property
+    def not_converted_yet(self):
+        return self.find_element(*self._ncy_locator).text
```
m1yag1: Your returning the text here. The return value of this is the text inside the locator. It is literally returning a string `NOT_CONVERTED_YET: caption`.

Your intuition on this is really good by the way but you don't want the text. You to return `is_element_displayed`.

pages/content.py
m1yag1: I also recommend not using the CSS selector but actually the By.CLASS_NAME in this case. I think CSS selectors should be a last resort really. I've used them so much b/c there has been no other way. In an ideal world we would have the devs add ids.

kerwinso:
I should clarify that the `NOT_CONVERTED_YET` test/fix only applies to the OpenStax American Government collection right now (that's why I changed the name of the test). The bug is still out there for other collections.

m1yag1:
Closing PR as it has been implemented in a different PR.

PR #7, opened 2018-04-17
ncy-example updates

kerwinso:
updates to the `ncy-example` branch


tests/utils.py
```diff
@@ -3,11 +3,12 @@
...
 
-    Assumes that each item is on a single line and strips newline characters
+    Assumes that each item is on a single line and strips leading and trailing characters
     """
-    with open(location, 'r') as f:
-        uuids = [uuid.strip() for uuid in f.readlines()]
-    return uuids
+    with open(filepath, 'r') as f:
+        for line in f:
+            line = line.strip()
```
m1yag1: we may be able to skip the assignment step and go directly to `return line.strip()`

kerwinso:
Finally rebased and force pushed.

Issue #11, opened 2018-04-19
tox blows up on relative import of the test data file


m1yag1:
This is probably due to the working directory of wherever the pytest command is being executed. The best bet is probably to create a variable in the `conftest.py` file that figures out it's own filepath and the relation of the data directory to that. 

I believe something like this may do the trick:

`DATA_DIR = os.path.join(os.path.realpath(os.path.dirname(__file__)), 'data')`

kerwinso:
Addressed by #12.

PR #12, opened 2018-04-19
Fixed relative import issue #11

m1yag1:
* Added a DATA_DIR variable to the conftest.py directory to figure out
where the data directory is in relation to itself.
* Used the DATA_DIR variable along with os.path.join() to join the data
directory and the test file


kerwinso:
After re-installing my venv and running `tox`, the tests look like they passed. But the flake8 stuff blew up. Summary is `ERROR:   flake8: commands failed`, but the actual errors ran thousands of lines.

PR #14, opened 2018-04-20
Add travis

m1yag1:


philschatz:
Is the plan to have Jenkins run docker instead of getting it the tests to run on Travis? Also, @pumazi mentioned that there is an Ubuntu docker image for Travis but I don't know how to set it up or what it is

m1yag1:
The tests currently run on Travis using docker. It was introduced in #17

Issue #10, opened 2018-04-19
Need to add a way to skip longer running tests like ncy


m1yag1:
Having a command line switch that skips long running tests for localdev would be great.

kerwinso:
fixed with #17

PR #17, opened 2018-04-25
add `--runslow` command line option to allow skipping ncy tests

kerwinso:
Addresses #10.

From the command line:
- `pytest` will skip `test_ncy_is_not_displayed` by default.
- `pytest --runslow` will include `test_ncy_is_not_displayed` (all 80-something tests).

In Pycharm:
- Add `--runslow` to Additional Arguments in your Configurations.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -16,7 +16,17 @@ def pytest_addoption(parser):
     group._addoption('--headless',
                      action='store_true',
                      help='enable headless mode for chrome.')
+    parser.addoption("--runslow", action="store_true",
+                     default=False, help="run slow tests")
 
+def pytest_collection_modifyitems(config, items):
+    if config.getoption("--runslow"):
+        # --runslow given in cli: do not skip slow tests
+        return
+    skip_slow = pytest.mark.skip(reason="need --runslow option to run")
+    for item in items:
+        if "slow" in item.keywords:
+            item.add_marker(skip_slow)
 
 @pytest.fixture
 def chrome_options(chrome_options, pytestconfig):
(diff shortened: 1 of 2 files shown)


tests/test_content.py
```diff
@@ -8,6 +8,7 @@
 
 
 @pytest.mark.nondestructive
+@pytest.mark.slow
```
m1yag1: Flip the order on these. It's a bit weird but decorators are executed in order from the bottom up. We want `@pytest.nondestructive` to be evaluated first.

tests/conftest.py
```diff
@@ -16,6 +16,18 @@ def pytest_addoption(parser):
     group._addoption('--headless',
                      action='store_true',
                      help='enable headless mode for chrome.')
+    parser.addoption('--runslow', action='store_true',
```
m1yag1: We can talk about this one. I'm wondering if maybe --runlong would be better. This test isn't really slow it just takes a long time to run because its a lot of tests for the same thing... we can always switch it out later if we write a test that is super slow. It's a little nitpicky.

kerwinso: Good point. But I'm leaning towards sticking with `--runslow` for a couple reasons:

1. It's more consistent with the syntax of the decorator (`@pytest.mark.slow`) as well as the pytest documentation. (Unless there's another decorator for `mark.long`? I don't think it would make sense to code one from scratch.)
1. Eventually we will probably need to apply `@pytest.mark.slow` to future tests that run _slow_, but not necessarily _long_. If we are applying the same decorator to both slow and long tests, wouldn't it be repeating ourselves to define 2 different options?

m1yag1: We can define our our marks for the decorator `pytest.mark.long`.  I think it makes more sense to keep going with slow according to your suggestions. Let's keep it that way. In effect, it is slowing down the throughput of the entire testing process. 👍

PR #21, opened 2018-05-07
Changed folder structure to support multiple cnx products (not just webview)

m1yag1:
* Moved webview tests into tests/cnx/functional
* Moved american govmt specific fixture into the test file.


m1yag1:
Closing b/c I went a different direction in another PR.

PR #25, opened 2018-05-11
(Maybe) fix stale element reference errors in webview tests

Dantemss:
This *seems* to work (passed on Travis 3x in a row) but you never know with these intermittent failures. The trick is checking that the elements found via CSS are actually descendants of the body element.


Dantemss:
Of course, the 4th time (when I actually make the PR) it fails.

PR #27, opened 2018-05-12
Fixed webview stale errors

m1yag1:
Implemented dante's body text approach and `super().loaded` call along with some other recommended items from the docs.


Dantemss:
I think all this needs is the `super().loaded` call in `Home` too and it should work.

Dantemss:
Oh maybe one more thing: I did convert the `find_elements` call in Home in my PR to use `is_element_present` instead and added a couple more selector checks from the other constants that are used in Home.

m1yag1:
Closing as this was mostly for exploration and my curiosity.

PR #31, opened 2018-05-15
Added legacy create/import/publish module tests

Dantemss:
Could potentially run the same tests on more modules or set other things like some metadata.


pages/archive/content.py
```diff
@@ -0,0 +1,96 @@
...
+        return ET.tostring(html, encoding='unicode')
+
+    @property
+    def stable_dict(self):
+        return {**{field: self.dict[field] for field in self._stable_fields},
+                **{'content': self.stable_content}}
+
+    @property
+    def loaded(self):
+        return super().loaded and self.is_element_displayed(*self._json_locator)
```
m1yag1: The page in `pages/archive/base.py` doesn't have a loaded property defined. Do you need to call `super().loaded` on that?

pages/archive/content.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # Remove revision time
+        revised_time = head.find('./{http://www.w3.org/1999/xhtml}meta[@name="revised-time"]')
+        head.remove(revised_time)
+
+        return ET.tostring(html, encoding='unicode')
+
+    @property
+    def stable_dict(self):
+        return {**{field: self.dict[field] for field in self._stable_fields},
```
m1yag1: Nice use of a python 3 only feature.

pages/archive/content.py
Dantemss: I could probably skip the call in this case (the default `loaded` method just returns true every time)

pages/legacy/module_edit.py
```diff
@@ -0,0 +1,103 @@
...
+    _publish_link_locator = (By.CSS_SELECTOR, 'a[href$="module_publish"]')
+
+    _import_form_locator = (By.CSS_SELECTOR, 'form[action="module_import_form"]')
+    _import_select_locator = (By.CSS_SELECTOR, 'select[name="format"]')
+
+    _content_textarea_locator = (By.ID, 'textarea')
+
+    @property
+    def username(self):
+        return re.search('/Members/([^/]+)/([^/]+)', self.driver.current_url).group(1)
```
m1yag1: You may want to add this regex as a property and use `re.compile` and then pass it into the `re.search()` function.
```python
_username_regex = re.compile('/Members/([^/]+)/([^/]+)')
@property
    def username(self):
        return re.search(self_username_regex, self.driver.current_url).group(1)
```

pages/archive/content.py
Dantemss: Thanks. Took a lot of googling and trying to figure out how to merge dicts properly :)

Dantemss:
@m1yag1 changes have been made

PR #36, opened 2018-05-15
Replace --headless option with --headed

Dantemss:
Since --headless is enabled by default, --headed will deactivate it.

Some random online thesaurus claimed headed was a real adjective and it fits so...


kerwinso:
Really, not '--headful'? 😛

m1yag1:
Headless mode is default? I remember adding the flag so that it ran headless b/c it wasn't the default. I'm confused ...

Dantemss:
Closing because we want headless to not be default

PR #57, opened 2018-05-16
Run cnx-automation natively in Travis containers (without Docker)

Dantemss:
Here it is, as promised.

Pros:
  - Faster builds (less time waiting for Travis to bootup, Docker and the image don't have to be installed, pip packages can be cached so they don't have to be downloaded every time)

Cons:
  - Travis no longer tests the Docker config

You can decide whether or not this tradeoff is worth it :)

This PR is based on the `headless` PR to avoid future conflicts and for consistency.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 python: 3.6
+cache: pip
 
 sudo: false
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
+python: 3.6
 
-sudo: required
+sudo: false
 
 env:
   global:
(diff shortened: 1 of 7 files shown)


m1yag1:
Dante, As much as I like this PR I'm trying to move QA projects to using docker. I've created a new PR #59 which pulls the `openstax/standalone-chrome-debug` docker image via our openstax dockerhub repository. I'd like to move more in this direction as this is the way the industry is going.

My long term goal is that when PRs are merged to master that triggers a docker build in travis to create an image that contains this repo and tags it with some version number and uploads it to dockerhub. Whenever we get Jenkins going, I would like Jenkins to receive a webhook from GitHub indicating any changes to webview, legacy, etc. and Jenkins pulls the `openstax/cnx-automation:<some_version>` image and runs all the necessary regression tests in order to notify the developer if there is a failure. Docker is heavily involved in my long term vision for how environments are configured for QA testing.

PR #60, opened 2018-05-18
Legacy collection tests

Dantemss:
Test creating a simple collection with 1 module and publishing it, then checking the content on CNX archive.


Dantemss:
I've seen at least 1 failure during login. Might be a good idea to figure out how to make Travis print the page source when a test fails.

pages/legacy/content_published.py
```diff
@@ -7,7 +7,7 @@
 from selenium.webdriver.common.by import By
 
 
-class PublishedModule(PrivatePage):
+class ContentPublished(PrivatePage):
     _tbody_locator = (By.CSS_SELECTOR, 'table.leftheadings tbody')
     _title_locator = (By.XPATH, ("./tr/th[text()='Name:']/following-sibling::td/span"))
```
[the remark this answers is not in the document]
Dantemss: Yeah I think matching on the text is literally the only thing I can do here to extract those attributes (well unless I do it by position of the tags but that's even more flaky).

pages/archive/content.py
```diff
@@ -35,38 +41,76 @@ class Content(Page):
...
+        """Returns whether or not the "tree" field is present."""
+        return 'tree' in self.dict
+
+    @property
+    def tree(self):
+        """Returns the archive tree json as a dict."""
+        return self.dict['tree']
+
+    def stable_tree(self, tree):
+        """Returns the archive tree json as a dict with fields that change
```
m1yag1: Your docstrings need to be changed to match the style guide.

Essentially, if you can write it on one line that's great but if it is two lines then it should be a very brief description followed by a new line and then a deeper description.

```python
# Good
def login():
"""Logs in.

Clicks the login link and then waits for the home page to load.

"""
```
```python
# Bad
def login():
"""Logs in.
Clicks the login link and then waits for the home page to load."""
```

pages/archive/content.py
```diff
@@ -85,5 +135,20 @@ def stable_content(self):
...
-                **{'content': self.stable_content}}
+        """Returns a stable version of the full archive json as a dict.
+
+        Returns from the archive json only fields that are
+        guaranteed not to change from test to test as a dict.
+
+        This includes the whitelisted fields in the _stable_fields array,
+        plus stable versions of the tree and content fields, if present.
+        """
+        dict = {field: self.dict[field] for field in self._stable_fields if field in self.dict}
```
m1yag1: dict is a built-in keyword in python. you may want to choose a different variable name.

regions/legacy/add_modules.py
```diff
@@ -0,0 +1,72 @@
...
+from selenium.webdriver.common.by import By
+
+from regions.legacy.base import Region
+
+
+class AddModules(Region):
+
+    _search_form_locator = (By.ID, 'collection-composer-collection-module-form-search')
+    _words_field_locator = (By.CSS_SELECTOR, 'input[type="text"][name="words"]')
+
```
m1yag1: this space is not necessary.

PR #64, opened 2018-05-22
Test that deletes all content from automation account's personal workspace

Dantemss:
Ross said this would be good to prevent the workspace from growing unbounded, potentially causing tests to take longer and longer.

Includes https://github.com/openstax/cnx-automation/pull/60, so probably merge that first, then rebase to master for smaller diff. Or just look at https://github.com/openstax/cnx-automation/pull/64/commits/46fa56826eaa577f08b2fb61edfd663018eda052 and https://github.com/openstax/cnx-automation/pull/64/commits/8f56b016448944ff1105e7ddc12b8935bc15779e for the diff after #60.

### DIFF
--- a/pages/legacy/base.py
+++ b/pages/legacy/base.py
@@ -8,12 +8,16 @@ from selenium.webdriver.common.by import By
 
 
 class Page(pypom.Page):
-    _region_content_locator = (By.ID, 'region-content')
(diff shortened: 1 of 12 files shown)


pages/legacy/base.py
```diff
@@ -8,13 +8,17 @@
 
 
 class Page(pypom.Page):
-    _region_content_locator = (By.ID, 'region-content')
```
Dantemss: Was unused

Issue #41, opened 2018-05-16
As a tester, I'd like a webview test that asserts Failed searches return 'No results found.'


m1yag1:
1. Go to test environment

2. Click on "search" on the top menu.

3. Type random query like "asdfafddafsf" into the search box, and hit return.

Dantemss:
https://github.com/openstax/cnx-automation/pull/67

Let me know if I should test for the Polish message too.

m1yag1:
Completed by #67

PR #67, opened 2018-05-22
Added webview test for searching and finding no results

Dantemss:
https://github.com/openstax/cnx-automation/issues/41

Let me know if I should test for the Polish message too.

### DIFF
--- a/pages/webview/browse.py
+++ b/pages/webview/browse.py
@@ -1,26 +1,35 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
 from pypom import Region
-from selenium.webdriver.common.by import By
 
 from pages.webview.base import Base
+
+from selenium.webdriver.common.by import By
 
 
 class Browse(Base):
     URL_TEMPLATE = '/browse'
 
-    _search_locator = (By.ID, 'find-content-input')
+    _search_input_locator = (By.ID, 'find-content-input')
+    _advanced_search_button_locator = (By.CLASS_NAME, 'advanced-search')
+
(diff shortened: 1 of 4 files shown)


pages/webview/search_results.py
```diff
@@ -0,0 +1,32 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from pages.webview.base import Base
+
+from selenium.webdriver.common.by import By
+
+
+class SearchResults(Base):
+    URL_TEMPLATE = '/search'
+
```
m1yag1: We don't need these spaced out.

pages/webview/browse.py
```diff
@@ -1,45 +1,61 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
 from pypom import Region
-from selenium.webdriver.common.by import By
```
m1yag1: This doesn't need to be changed but a lot of the imports are incorrect. The order of imports should be:
1. standard lib
2. libraries
3. application modules

Sooo. selenium should be above any pages that are imports. This doesn't need to be changed right now. We can do it later.

m1yag1: a lot of that is my fault =(

tests/webview/ui/test_browse.py
```diff
@@ -31,7 +32,7 @@ def test_search_no_results(base_url, selenium):
     # GIVEN the browse page and a bogus query
     home = Home(selenium, base_url).open()
     browse = home.header.click_browse()
-    query = token_urlsafe(32)
+    query = ''.join(choice(digits + ascii_letters) for i in range(32))
```
m1yag1: You can use the python3.6 module now if you like.

PR #137, opened 2018-06-18
Add WEBVIEW_BASE_URL env var and move Travis back to QA

Dantemss:
- Renamed base_url to webview_base_url
- Restored functionality that skips destructive tests on sensitive urls

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,9 +1,9 @@
 [tool:pytest]
 addopts=-vs --driver=Chrome --html report.html
-base_url=https://staging.cnx.org
 archive_base_url=https://archive-qa.cnx.org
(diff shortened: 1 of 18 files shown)


setup.cfg
```diff
@@ -1,9 +1,9 @@
 [tool:pytest]
 addopts=-vs --driver=Chrome --html report.html
-base_url=https://staging.cnx.org
 archive_base_url=https://archive-qa.cnx.org
 legacy_base_url=https://legacy-qa.cnx.org
-sensitive_url=^(?:https?://)?cnx\.org
+webview_base_url=https://staging.cnx.org
```
m1yag1: So are we saying that local runs should hit staging and travis should hit qa?

Issue #144, opened 2018-06-19
AAT, I'd like a webview test that asserts Top navigation works in both desktop and responsive mode


m1yag1:
# CASE NUM: C167405

## User story

```
AAT, I'd like a webview test that asserts Top navigation works in both desktop and responsive mode
```

## Preconditions

```

```

## Steps

```
Check for the following elements in top nav, that they are there and they work:

Top right:

- Support goes to http://openstax.force.com/support?l=en_US&c=Products%3ACNX
- CNX Author Legacy Site link goes to https://legacy-staging.cnx.org/content?legacy=true


Top nav:

- OpenStax CNX logo goes to https://staging.cnx.org/
- Search goes to https://staging.cnx.org/browse
- About Us goes to https://staging.cnx.org/about
- Give goes to https://staging.cnx.org/donate
- Rice logo goes to rice.edu

When reducing browser width (responsive mode), the OpenStax CNX logo, Support, and Legacy Site links should remain visible, and the other nav items should be found within a hamburger menu.
```

## Expected Result

```

```

Dantemss:
https://github.com/openstax/cnx-automation/pull/171

PR #171, opened 2018-06-19
Added more webview nav tests

Dantemss:
https://github.com/openstax/cnx-automation/issues/144

### DIFF
--- a/tests/webview/ui/test_home.py
+++ b/tests/webview/ui/test_home.py
@@ -14,6 +14,7 @@ _number_of_tested_books = 2
 
 
 @markers.webview
+@markers.test_case('C167405')
 @markers.nondestructive
 @markers.parametrize('width,height', [(1024, 768), (640, 480)])
 def test_top_right_links_and_nav(width, height, webview_base_url, legacy_base_url, selenium):
--- a/tests/webview/ui/test_home.py
+++ b/tests/webview/ui/test_home.py
@@ -15,26 +15,61 @@ _number_of_tested_books = 2
 
 @markers.webview
 @markers.nondestructive
-def test_splash_banner_loads(webview_base_url, selenium):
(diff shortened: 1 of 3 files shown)


tests/webview/ui/test_home.py
```diff
@@ -15,26 +15,61 @@
 
 @markers.webview
 @markers.nondestructive
-def test_splash_banner_loads(webview_base_url, selenium):
-    # GIVEN the main website URL and the Selenium driver
+@markers.parametrize('width,height', [(1024, 768), (640, 480)])
```
m1yag1: We could also add this to the selenium fixture which would run all the tests with both resolutions.

m1yag1:
```python
@pytest.fixture(scope='function',
                params=[(1080, 1920), (414, 738)],
                ids=['Resolution: 1080x1920', 'Resolution: 414x738'])
def selenium(selenium, request):
    """Fixture to set custom selenium parameters.
    This fixture will also parametrize all of the tests to run them on both a
    Desktop resolution and a mobile resolution.
    Desktop size: 1920x1080
    Mobile size: 738x414 (iPhone 7+)
    """
    selenium.set_window_size(*request.param)
    return selenium
```

m1yag1: We can hold off on this for now. It will make things take longer and I don't think the expectations of webview are high for responsiveness. We can pick out select tests like the hamburger nav to test and that's good 👍

tests/webview/ui/test_home.py
```diff
@@ -15,26 +15,61 @@
 
 @markers.webview
```
m1yag1: needs @markers.test_case('<case_id>')

Dantemss: Added

Issue #146, opened 2018-06-19
AAT, I'd like a webview test that asserts HTML markup should not be visible in section titles


m1yag1:
# CASE NUM: C167408

## User story

```
AAT, I'd like a webview test that asserts HTML markup should not be visible in section titles
```

## Preconditions

```

```

## Steps

```
Look at any title of any section.
```

## Expected Result

```
It should read something like `1.4 The Language of Physics`

NOT

`<span class="os number">1.4</span><span class="os text">The Language of Physics</span>`
```

Dantemss:
https://github.com/openstax/cnx-automation/pull/175

m1yag1:
Resolved by #175

PR #175, opened 2018-06-20
Test that the section title does not contain HTML markup

Dantemss:
https://github.com/openstax/cnx-automation/issues/146

### DIFF
--- a/tests/webview/ui/test_content.py
+++ b/tests/webview/ui/test_content.py
@@ -39,8 +39,9 @@ def test_navs_and_elements_are_displayed(webview_base_url, selenium):
 
 
 @markers.webview
+@markers.test_case('C132542')
 @markers.nondestructive
-def test_author(webview_base_url, selenium):
+def test_author_is_openstax(webview_base_url, selenium):
     # GIVEN the home page and a book
     home = Home(selenium, webview_base_url).open()
     book = home.featured_books.openstax_list[0]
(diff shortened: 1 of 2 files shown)


tests/webview/ui/test_content.py
```diff
@@ -163,6 +163,22 @@ def test_get_this_book(webview_base_url, selenium):
     assert offline_zip_displayed == downloads.is_offline_zip_available
 
 
+@markers.webview
```
m1yag1: Add `@markers.test_case()`

tests/webview/ui/test_content.py
```diff
@@ -163,6 +163,22 @@ def test_get_this_book(webview_base_url, selenium):
...
+@markers.nondestructive
+def test_section_title(webview_base_url, selenium):
```
m1yag1: Be more specific with the test name. `test_section_title_for_no_markup`

tests/webview/ui/test_content.py
```diff
@@ -163,6 +163,22 @@ def test_get_this_book(webview_base_url, selenium):
...
+
+    # THEN the section title does not contain HTML markup
+    section_title = content.section_title
+    assert '<' not in section_title
+    assert '>' not in section_title
+
+
 @markers.webview
 @markers.nondestructive
 def test_content(webview_base_url, selenium):
```
m1yag1: This also looks like it needs a more specific name and testrail marker

m1yag1: oh snap, that's from the old import where i didn't include that, my badz. @kerwinso can you help dante locate the case id for this one?

tests/webview/ui/test_content.py
Dantemss: Added

tests/webview/ui/test_content.py
Dantemss: Changed the test name to `test_section_title_for_no_markup`

PR #178, opened 2018-06-20
Added test rail integration

m1yag1:
* added testrail.example.cfg
* added test case id to featured books test
* added testrail marker to markers.py
* added testrail.cfg to .gitignore


m1yag1:
Oh good point! I'll add testrail info into the README. i meant to do that but forgot.

PR #198, opened 2018-06-28
DO NOT MERGE: trying out applitools

m1yag1:
This is only here while I play around with applitools.


m1yag1:
Closing this out for now. This is less of a focus at the moment.

Issue #158, opened 2018-06-19
AAT, I'd like a webview test that asserts Long search query should not break search


m1yag1:
# CASE NUM: C175151

## User story

```
AAT, I'd like a webview test that asserts Long search query should not break search
```

## Preconditions

```

```

## Steps

```
Load this link:
https://staging.cnx.org/search?q=subject:%22Science%20and%20Technology%22%20pubYear:%222013%22%20type:%22Book%22%20physics%20introduction%20examples%20college%20online%20learning%20and%20algebra
```

## Expected Result

```
Search results should return something around 2-3 Books. If it returns a message "Search took too long", reload the browser and the search should work.

If zero search results are returned, this test fails.
```

Dantemss:
This query caused a timeout for me on QA.

Dantemss:
I will add code to retry the search if it times out.

Dantemss:
https://github.com/openstax/cnx-automation/pull/203

PR #203, opened 2018-06-29
Test long searches

Dantemss:
- Split test_browse into test_browse and test_search
- Added code to retry searches that timeout
- Added test using a long search string

#158


Dantemss:
https://github.com/openstax/cnx-automation/issues/158

Issue #169, opened 2018-06-19
AAT, I'd like a webview test that asserts Top nav links should have new ids in HTML


m1yag1:
# CASE NUM: C195236

## User story

```
AAT, I'd like a webview test that asserts Top nav links should have new ids in HTML
```

## Preconditions

```

```

## Steps

1. go to front page of testing environment
1. in your browser, inspect the element classes for the top navigation links

## Expected Result

- Search link should have `id="nav-browse"`
- About link should have `id="nav-about"`
- Give link should have `id="nav-donate"`

Dantemss:
https://github.com/openstax/cnx-automation/pull/213

See comment in the PR about maybe reversing the test case.

Dantemss:
Closing because it was decided we do not need to explicitly test for these.

PR #213, opened 2018-07-03
Test the ids of top nav links

Dantemss:
https://github.com/openstax/cnx-automation/issues/169

I could instead reverse the test case and select by ID as usual and just test the text of the links instead of using these crazy XPATHs.

Based on https://github.com/openstax/cnx-automation/pull/205 to avoid conflicts. Look at only the last commit for diff.


Dantemss:
Closing because it was decided we do not need to explicitly test for these.

Issue #233, opened 2018-07-09
AAT, I'd like a webview test that asserts buggy-looking loading message should not appear in left pane


m1yag1:
# CASE NUM: C195071
## Section

Show Which Books a Single Page is in

https://trello.com/c/i560M3Af/422-front-end-show-which-books-a-single-page-is-in
Bugs: https://github.com/Connexions/webview/milestone/3

## User story

AAT, I'd like a webview test that asserts buggy-looking loading message should not appear in left pane

## Steps

1. Go to this link https://staging.cnx.org/contents/yWshIYVW/How-Changes-in-the-Cost-of-Liv
1. Watch the left pane closely.

## Expected Result

As the page loads, there should _not_ be a message 'This page is in books' in the left pane without any links to books. If there is, it's a regression.

m1yag1:
@kerwinso A quick question: the bug in this test case is when there are no books listed right?

kerwinso:
@m1yag1 @BrianYing  In this case, there _should_ be a book listed in the left pane, because this page does belong to a book.

This is how it should look:
![image](https://user-images.githubusercontent.com/11298165/42580907-37cfb510-84f1-11e8-8a75-eb7fe93fdfce.png)

 

The bug is that in the left pane, the link to the book was taking a long time to load, and in the meantime the left pane was also displaying a 'This page is in books' message (message varies depending on how many books the page is in). 

It would look something like this (which looks like a bug):
![image](https://user-images.githubusercontent.com/11298165/42580831-0ac4838e-84f1-11e8-9ee5-6dc2c87c7684.png)

So I guess the automation would have to check that both the 'This page is in books' message (or equivalent) does not display without link(s) to the books.

m1yag1:
@kerwinso ok so i'm reviewing the test case and this seems to specifically be when the page is loading. That may not be something we can check. All our tests wait for the page to be loaded before doing anything. This may not be something we can automate.

kerwinso:
@m1yag1 That's OK if we can't automate it. It's not a high-priority regression-- just a less-than-optimized user experience that I haven't really been able to reproduce anyway (the screenshot above was Photoshopped based on my memory). 😃

kerwinso:
BTW, here's the original github issue: https://github.com/Connexions/webview/issues/1620

m1yag1:
this is already covered by #245

m1yag1:
This issue is closed to it being very difficult to replicate during automation. It also isn't a high priority regression.

PR #236, opened 2018-07-10
Added test case for issue 233 with the buggy stuff on the left panel

BrianYing:
#233


tests/webview/ui/test_content.py
```diff
@@ -766,3 +766,18 @@ def test_chapter_review_version_matches_book_version(webview_base_url, selenium,
 
     # THEN the chapter review version matches the book version
     assert content.page_version == content.book_version
+
+
+@markers.webview
+@markers.test_case('C195071')
+@markers.nondestructive
+@markers.parametrize('id', ['yWshIYVW@1'])
+def test_buggy_msg_not_show_in_left_panel(webview_base_url, id, selenium):
```
m1yag1: I realize the name of the test included buggy load message but we also want to take a second to reflect on exactly what we are testing here. Remember: we are about quality in all things ;). As I read the test case what is REALLY being tested is that when the `This page is in books` message exists in the left pane that it has books listed underneath it.

Please rename the test to reflect what your are testing. A better name may be:
`test_this_page_in_books_contains_books`

tests/webview/ui/test_content.py
```diff
@@ -766,3 +766,18 @@ def test_chapter_review_version_matches_book_version(webview_base_url, selenium,
...
+    # GIVEN the webview base url, a chapter page id, and the Selenium driver
+
+    # WHEN we visit that page of the chapter
+    content = Content(selenium, webview_base_url, id=id).open()
+
+    # THEN correct book info is listed in the left panel
```
m1yag1: You have the idea down of using the POM and executing methods for the test but what happens here if `is_left_panel_title_displayed` is false?

The answer is the test passes. Is that what we want?

pages/webview/content_page.py
```diff
@@ -23,14 +22,31 @@ def books_containing(self):
...
         class Book(Region):
             _title_locator = (By.CSS_SELECTOR, 'div')
+            _go_to_book_locator = (By.CSS_SELECTOR, 'ul > li > div')
 
             @property
             def title(self):
                 return self.find_element(*self._title_locator).text
+
+            @property
+            def go_to_book(self):
```
m1yag1: rename `go_to_book_link`

pages/webview/content.py
```diff
@@ -31,6 +31,8 @@ class Content(Page):
                           ' button[type="submit"][data-l10n-id="textbook-view-btn-get-this-book"]')
     )
     _ncy_locator = (By.CLASS_NAME, 'not-converted-yet')
+    _book_info_title_locator = (By.CSS_SELECTOR, '.booksContaining > div > span')
```
m1yag1: These should be removed. They are already on `pages/content_page.py`

pages/webview/content.py
```diff
@@ -114,6 +116,14 @@ def get_this_book_button(self):
     def is_ncy_displayed(self):
         return self.is_element_displayed(*self._ncy_locator)
 
+    @property
```
m1yag1: These also need to be removed

pages/webview/content_page.py
```diff
@@ -23,14 +22,31 @@ def books_containing(self):
     class BooksContaining(Region):
         _root_locator = (By.CLASS_NAME, 'booksContaining')
         _book_list_locator = (By.CSS_SELECTOR, 'div > ul > li')
+        _nav_title_locator = (By.CSS_SELECTOR, 'span.title')
```
m1yag1: This is called the `_overview_locator`

pages/webview/content_page.py
```diff
@@ -23,14 +22,31 @@ def books_containing(self):
...
+            return self.find_element(*self._nav_title_locator)
 
         @property
+        @retry_stale_element_reference_exception
         def book_list(self):
             return [self.Book(self.page, el) for el in self.find_elements(*self._book_list_locator)]
 
+        @property
+        @retry_stale_element_reference_exception
+        def go_to_book_link(self):
```
m1yag1: This is called `go_to_book_link` but returns a list of links? I don't think this is necessary. In your test you should iterate over the book list and assert that there is a `got_to_book_link`.

```
book_list = content.books_containing.book_list

for book in book_list:
    assert book.go_to_book_link.is_displayed()

File: pages/webview/content_page.py [hunk 8]
```diff
@@ -23,14 +22,31 @@ def books_containing(self):
...
+            return [self.Book(self.page, el).go_to_book for el in
+                    self.find_elements(*self._book_list_locator)]
+
         class Book(Region):
             _title_locator = (By.CSS_SELECTOR, 'div')
+            _go_to_book_locator = (By.CSS_SELECTOR, 'ul > li > div')
```
m1yag1: rename `_go_to_book_link_locator`

Issue #228, opened 2018-07-09
AAT, I'd like a webview test that asserts Book title link should not be limited in length


m1yag1:
# CASE NUM: C195061
## Section

Show Which Books a Single Page is in

https://trello.com/c/i560M3Af/422-front-end-show-which-books-a-single-page-is-in
Bugs: https://github.com/Connexions/webview/milestone/3

## User story

AAT, I'd like a webview test that asserts Book title link should not be limited in length

## Steps

1. Go to this link https://staging.cnx.org/contents/BWYBGK7C@2/Executive-Summary
1. Look at the title in the left nav, it should say: 
`"Our Cultural Commonwealth" The Report of the American Council of Learned Societies Commission on Cyberinfrastructure for the Humanities and Social Sciences`

## Expected Result

The title in the left nav should not be cut off or limited to 2 lines. You should be able to read the full title of the book which runs 5 lines long.

m1yag1:
Hey @kerwinso , we need a little more context on this issue. When the text was cut-off was it completely cut-off as in the text wasn't there or was it obscured by another element? The reason I ask is that the test wants to make sure long lines of text are world-wrapped appropriately. There is no way to detect that in selenium. When we use the `.text` attribute it returns the text only.

kerwinso:
@m1yag1 This test originated from this issue: https://github.com/Connexions/webview/issues/1610

When the text was cut off, it was cut off at the second line's end point with an ellipsis `...` and then the rest of the book title would not display. 

**Example:** 
`Programming Fundamentals - A Modular Structured Approach using C++`
became
`Programming Fundamentals - A Modular Structured Approach using...`

Like in this screenshot: https://user-images.githubusercontent.com/253212/38148312-867c56d6-341b-11e8-9e03-fa2580bd34eb.png

This was originally implemented as an aesthetic/design choice. However, we decided to remove the title length limitation altogether, because some important info was being lost in the truncated titles.

So in brief, the acceptance criteria is that the full title of the book should display in the browser with no truncation at all. (I'm not so worried about word-wrapping.) 

I think if you are able to grab the full title in the `.text` attribute in selenium for [the book specified in the test](https://staging.cnx.org/contents/BWYBGK7C@2/Executive-Summary), that should be fine to consider as passing the test.

**Example:**
- `link.text = "Our Cultural Commonwealth" The Report of the American Council of Learned Societies Commission on Cyberinfrastructure for the Humanities and Social Sciences` == PASS
[...]

PR #238, opened 2018-07-11
Book title length limitation

BrianYing:
#228

### DIFF
--- a/tests/webview/ui/test_content.py
+++ b/tests/webview/ui/test_content.py
@@ -792,10 +792,7 @@ def test_book_title_not_limited(webview_base_url, id, selenium):
 
     # WHEN we visit that page of the chapter
     content = Content(selenium, webview_base_url, id=id).open()
-
+    sleep(0.25)
     # THEN get the title text to check if it matches what it suppose to be
     actual_title = content.get_left_nav_book_title
-    correct_title = '"Our Cultural Commonwealth" The Report of the American ' \
-                    'Council of Learned Societies Commission on Cyberinfrastructure ' \
-                    'for the Humanities and Social Sciences'
-    assert actual_title == correct_title
+    assert "..." not in actual_title
--- a/tests/webview/ui/test_content.py
+++ b/tests/webview/ui/test_content.py
@@ -766,3 +766,17 @@ def test_chapter_review_version_matches_book_version(webview_base_url, selenium,
 
     # THEN the chapter review version matches the book version
     assert content.page_version == content.book_version
+
+
+@markers.webview
+@markers.test_case('C195071')
+@markers.nondestructive
+@markers.parametrize('id', ['yWshIYVW@1'])
+def test_buggy_msg_not_show_in_left_panel(webview_base_url, id, selenium):
+    # GIVEN the webview base url, a chapter page id, and the Selenium driver
+
+    # WHEN we visit that page of the chapter
+    content = Content(selenium, webview_base_url, id=id).open()
+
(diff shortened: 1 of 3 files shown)


m1yag1:
@BrianYing we may need to hold off on this one for bit. I'm gathering info from Kerwin on the issue itself. The ability to make sure text is word-wrapped correctly is not currently a capability that exists using Selenium.

tests/webview/ui/test_content.py
```diff
@@ -766,3 +766,36 @@ def test_chapter_review_version_matches_book_version(webview_base_url, selenium,
...
+@markers.nondestructive
+@markers.parametrize('id', ['BWYBGK7C@2'])
+def test_book_title_not_limited(webview_base_url, id, selenium):
+    # GIVEN the webview base url, a chapter page id, and the Selenium driver
+
+    # WHEN we visit that page of the chapter
+    content = Content(selenium, webview_base_url, id=id).open()
+
+    # THEN get the title text to check if it matches what it suppose to be
+    actual_title = content.get_left_nav_book_title
```
m1yag1: the `get_left_nav_book_title` is ok but when we get text we simply reference the text attribute. Change this to work in this fashion:
`book_title = content.left_nav_book_title.text`

tests/webview/ui/test_content.py
```diff
@@ -766,3 +766,36 @@ def test_chapter_review_version_matches_book_version(webview_base_url, selenium,
...
+    correct_title = '"Our Cultural Commonwealth" The Report of the American ' \
+                    'Council of Learned Societies Commission on Cyberinfrastructure ' \
+                    'for the Humanities and Social Sciences'
+    assert actual_title == correct_title
```
m1yag1: We do not want a hardcoded title. When something is hardcoded in ANY code you always want to question what you are doing. The test case does suggest that what you want to do is compare the title text but as revealed in the discussion #228 what we actually want to is to test that the ellipses `...` is not in the title.

I believe that would look something like this:
`assert '...' not in book_title`

tests/webview/ui/test_content.py
```diff
@@ -766,3 +766,33 @@ def test_chapter_review_version_matches_book_version(webview_base_url, selenium,
...
+@markers.nondestructive
+@markers.parametrize('id', ['yWshIYVW@1'])
+def test_buggy_msg_not_show_in_left_panel(webview_base_url, id, selenium):
+    # GIVEN the webview base url, a chapter page id, and the Selenium driver
+
+    # WHEN we visit that page of the chapter
+    content = Content(selenium, webview_base_url, id=id).open()
+
+    # THEN correct book info is listed in the left panel
+    if content.is_left_panel_title_displayed:
```
m1yag1: What happens if `content.is_left_panel_title_displayed == false`?

tests/webview/ui/test_content.py
```diff
@@ -766,3 +766,33 @@ def test_chapter_review_version_matches_book_version(webview_base_url, selenium,
...
+@markers.webview
+@markers.test_case('C195061')
+@markers.nondestructive
+@markers.parametrize('id', ['BWYBGK7C@2'])
+def test_book_title_not_limited(webview_base_url, id, selenium):
+    # GIVEN the webview base url, a chapter page id, and the Selenium driver
+
+    # WHEN we visit that page of the chapter
+    content = Content(selenium, webview_base_url, id=id).open()
+    sleep(0.25)
```
m1yag1: avoid sleeps plz.

Issue #221, opened 2018-07-09
AAT, I'd like a webview test that asserts The "Contents" button (which normally controls opening and closing the TOC) is now labelled "Books" and controls opening and closing this list.


m1yag1:
# CASE NUM: C195055
## Section

Show Which Books a Single Page is in

https://trello.com/c/i560M3Af/422-front-end-show-which-books-a-single-page-is-in
Bugs: https://github.com/Connexions/webview/milestone/3

## User story

AAT, I'd like a webview test that asserts The "Contents" button (which normally controls opening and closing the TOC) is now labelled "Books" and controls opening and closing this list.

## Steps

Can use this link to test: https://staging.cnx.org/contents/4fGVMb7P/Taboo-language-and-slanguage

m1yag1:
@BrianYing you have half of this done in #248 but you need to finish by closing and opening the books nav via the button.

kerwinso:
@m1yag1 Isn't this handled by https://github.com/openstax/cnx-automation/blob/a80a74fbb4c0d67dae76e1074b07bba868de76b8/tests/webview/ui/test_content.py#L897-L909 ? I just ran cnx-automation on staging and it passed this test in TestRail: https://openstax.testrail.com/index.php?/tests/view/346445

m1yag1:
Yea but it is not complete. it makes sure the `btn_name == 'books'` but doesn't click the button to ensure it opens and closes.

m1yag1:
@kerwinso so that it is more accurate for the time being you could rename the test in test rail to something that ensures the button is labelled books and create a new test case for the side bar to open and close. that should provide a better result. whatever you want to do. You could also remove the case id from the framework and it wouldn't mark it as passed.

PR #248, opened 2018-07-19
add test case for the button text

BrianYing:
#221


Dantemss:
Actually I just had a thought while reading the test case... Maybe this should click the button and test that the list opens/closes?

m1yag1:
Wait this doesn't close and open the list ...

Issue #220, opened 2018-07-09
AAT, I'd like a webview test that asserts "Books this page is in" list appears where the Table of Contents (nav sidebar) would appear in a book.


m1yag1:
# CASE NUM: C195054
## Section

Show Which Books a Single Page is in

https://trello.com/c/i560M3Af/422-front-end-show-which-books-a-single-page-is-in
Bugs: https://github.com/Connexions/webview/milestone/3

## User story

AAT, I'd like a webview test that asserts "Books this page is in" list appears where the Table of Contents (nav sidebar) would appear in a book.

## Steps

When loading a page outside the context of a book (e.g., https://staging.cnx.org/contents/4fGVMb7P/Taboo-language-and-slanguage ), a new "Books this page is in" list appears where the Table of Contents (nav sidebar) would appear in a book.

m1yag1:
This is closed by #249

PR #249, opened 2018-07-19
add test case to check the existence of "books this page is in" list

BrianYing:
#220


tests/utils.py
```diff
@@ -71,3 +73,33 @@ def skip_if_destructive_and_sensitive(request, base_url):
...
+        raise IndexError('Unknown dimension: %s' % dimension)
+    return get_size[dimension]
+
+
+def set_window_size(self, width=0, height=0):
+    """Attempt to change the browser window size."""
+    if width >= 1 and height >= 1:
+        self.driver.set_window_size(width, height)
+        sleep(1.0)
+    # return self.get_window_size()
```
m1yag1: Do not merge code with commented code. If there is commented code it either needs to be deleted or be necessary.

tests/utils.py
```diff
@@ -71,3 +73,33 @@ def skip_if_destructive_and_sensitive(request, base_url):
 def shorten_tag(tag):
     """Returns the short version of a git tag when given the long (or short) version."""
     return tag.split('-')[0]
+
+
+def get_window_size(self, dimension=None):
```
m1yag1: functions outside of a class with a reference to self are not good. You'll need to place these within a class. You could add them to the base page and they would be available everywhere.

tests/webview/ui/test_content.py
```diff
@@ -784,3 +786,35 @@ def test_book_containing_title_not_limited(webview_base_url, selenium, page_id):
...
+    # GIVEN the webview base url, page_id, and the Selenium driver
+
+    # WHEN we visit that page of the chapter and we have a list of books containing the page
+    content = ContentPage(selenium, webview_base_url, id=page_id).open()
+
+    # THEN check if the books list exists and on the left
+    assert content.books_containing.book_list
+
+    window_width = utils.get_window_size(content, 'width')
+    assert content.location['x'] < window_width/2
```
m1yag1: would `selenium.get_window_size()['width'] / 2`  not work. Additionally, you should be placing a space around operators in an expression such as in my example.

pages/webview/base.py
```diff
@@ -74,6 +76,31 @@ def offscreen_click(self, element):
...
+            raise IndexError('Unknown dimension: %s' % dimension)
+        return get_size[dimension]
+
+    def set_window_size(self, width=0, height=0):
+        """Attempt to change the browser window size."""
+        if width >= 1 and height >= 1:
+            self.driver.set_window_size(width, height)
+            sleep(1.0)
+
+    def set_window_position(self, x_=0, y_=0):
```
m1yag1: Why `x_` instead of `x`?

tests/webview/ui/test_content.py
```diff
@@ -816,5 +814,5 @@ def test_books_this_page_in_list(webview_base_url, selenium, page_id):
     # THEN check if the books list exists and on the left
     assert content.books_containing.book_list
 
-    window_width = utils.get_window_size(content, 'width')
-    assert content.location['x'] < window_width/2
+    window_width = content.get_window_size('width')
```
m1yag1: move to under the `WHEN` clause and edit WHEN comment.

`# WHEN we load the page of the chapter and we have the width of the window`

pages/webview/content_page.py
```diff
@@ -11,14 +13,26 @@ def loaded(self):
...
+    @property
+    def location(self):
+        content = self.find_element(By.CSS_SELECTOR, ".table-of-contents")
+        return content.location
+
     class BooksContaining(Region):
         _root_locator = (By.CLASS_NAME, 'booksContaining')
         _book_list_locator = (By.CSS_SELECTOR, 'div > ul > li')
 
+        # @property
```
m1yag1: Don't leave comments in code you want merged.

tests/webview/ui/test_content.py
```diff
@@ -784,3 +784,35 @@ def test_book_containing_title_not_limited(webview_base_url, selenium, page_id):
...
+
+
+@markers.webview
+@markers.test_case('C195054')
+@markers.nondestructive
+@markers.parametrize('page_id', ['4fGVMb7P@1'])
+def test_books_this_page_in_list(webview_base_url, selenium, page_id):
+    # GIVEN the webview base url, page_id, and the Selenium driver
+
+    # WHEN we visit that page of the chapter and we have a list of books containing the page
```
m1yag1: Comment is not exactly accurate... you don't have a list of books containing the page.

pages/webview/base.py
BrianYing: Honestly, I don't know. This are some old codes that @gregfitch wrote for automation and I just copied.

tests/webview/ui/test_content.py
```diff
@@ -784,3 +784,34 @@ def test_book_containing_title_not_limited(webview_base_url, selenium, page_id):
...
+    # THEN the button name is "Books" instead of "Contents"
+    btn_name = content.header_nav.contents_button.text
+    assert btn_name == "Books"
+
+
+@markers.webview
+@markers.test_case('C195054')
+@markers.nondestructive
+@markers.parametrize('page_id', ['4fGVMb7P@1'])
+def test_books_this_page_in_list(webview_base_url, selenium, page_id):
```
m1yag1: I missed this last time. This test needs to be renamed to something like `test_books_containing_list_is_on_left_of_page`

Issue #222, opened 2018-07-09
AAT, I'd like a webview test that asserts This list starts *open* (unlike the TOC) unless the initial page width is < 640 (i.e. mobile)


m1yag1:
# CASE NUM: C195056
## Section

Show Which Books a Single Page is in

https://trello.com/c/i560M3Af/422-front-end-show-which-books-a-single-page-is-in
Bugs: https://github.com/Connexions/webview/milestone/3

## User story

AAT, I'd like a webview test that asserts This list starts *open* (unlike the TOC) unless the initial page width is < 640 (i.e. mobile)

## Steps

Can use this link to test: https://staging.cnx.org/contents/4fGVMb7P/Taboo-language-and-slanguage

be sure to check the mobile width (< 640 px wide) as well -- that width will have the books list default closed

kerwinso:
@BrianYing found that if the width is exactly 640px, then the books list is open but the contents do not appear. This was already reported in https://github.com/Connexions/webview/issues/1626.

m1yag1:
This is closed by #250

PR #250, opened 2018-07-19
Add test case for window size shrink

BrianYing:
#222


tests/webview/ui/test_content.py
```diff
@@ -924,3 +925,24 @@ def test_books_containing_list_is_on_left_of_page(webview_base_url, selenium, pa
...
+@markers.parametrize('page_id', ['4fGVMb7P@1'])
+def test_button_open_with_certain_window_size(webview_base_url, selenium, page_id):
+    # GIVEN the webview base url, page_id, and the Selenium driver
+
+    # WHEN we visit that page of the chapter and we have a list of books containing the page
+    content = ContentPage(selenium, webview_base_url, id=page_id).open()
+
+    # THEN if window width >= 640, button should be open
+    window_width = content.get_window_size('width')
+    if window_width >= 640:
```
m1yag1: Let's not hardcode things. you can pass in the window dimensions via parametrize.

https://github.com/openstax/cnx-automation/blob/master/tests/webview/ui/test_home.py#L19

PR #252, opened 2018-07-23
Updates for the new webview home page

Dantemss:
- Rewrote the Read More test.
- Removed the `xfail` from the book intros test because all OpenStax books now have intros and CNX books are basically a WON'T FIX because it would require us to contact the authors to get the intro text.
- Made the scrolling tests scroll farther down since Prealgebra has less intro page content.
- Retry StaleElementReferenceExceptions while waiting for the ContentPage to load.
- Fixed the legacy module publish test by removing the data-cnxml-to-html-ver attribute from the module's content body tag.


Dantemss:
The change in the snapshot is simply the removal of `data-cnxml-to-html-ver=\"1.3.1+5.g23b5afc\"`

Issue #227, opened 2018-07-09
AAT, I'd like a webview test that asserts The book title should link to the book and navigate directly to the specific page in the book you were viewing


m1yag1:
# CASE NUM: C195060
## Section

Show Which Books a Single Page is in

https://trello.com/c/i560M3Af/422-front-end-show-which-books-a-single-page-is-in
Bugs: https://github.com/Connexions/webview/milestone/3

## User story

AAT, I'd like a webview test that asserts The book title should link to the book and navigate directly to the specific page in the book you were viewing

## Steps

1. go to https://staging.cnx.org/contents/4fGVMb7P@1/Taboo-language-and-slanguage
1. in left nav, click the first instance of the link:`English Home Language Grade 5`. It should take you to that page in the context of that book.
1. in left nav, click Contents to view the table of contents.

## Expected Result

The page 'Taboo language and slanguage' in the left nav should be highlighted green, showing you which page you are in inside the context of the book.

m1yag1:
This is closed by #255

PR #255, opened 2018-07-25
Title link and highlight viewing 2

BrianYing:


tests/webview/ui/test_content.py
```diff
@@ -799,3 +799,26 @@ def test_toc_button_labelled_books(webview_base_url, selenium, page_id):
...
+    # THEN click the title
+    content.get_left_nav_book_title.click()
+
+    # AND get and click the Contents button
+    content.header_nav.click_contents_button()
+
+    # AND find the on viewing title and get the color
+    chapter_4 = content.header_nav.table_of_contents.chapters[-1]
+    page_6_color = chapter_4.pages[-1].color()
+    assert page_6_color == right_color
```
m1yag1: Where is the THEN statements. 

Remember all asserts should be under THEN.

pages/webview/content.py
```diff
@@ -31,6 +32,10 @@ class Content(Page):
                           ' button[type="submit"][data-l10n-id="textbook-view-btn-get-this-book"]')
     )
     _ncy_locator = (By.CLASS_NAME, 'not-converted-yet')
+    _book_info_title_locator = (By.CSS_SELECTOR, '.booksContaining > div > span')
```
m1yag1: Books containing is specific to a different page. If you need access to `booksContaining` then you must use `pages/content_page.py` which represents a single page all by itself w/o a book.

pages/webview/content.py
```diff
@@ -31,6 +32,10 @@ class Content(Page):
...
+    _go_to_book_link_locator = (By.CSS_SELECTOR, 'li:nth-child(3) > div > a')
+    _left_nav_book_title_locator = (By.CSS_SELECTOR, 'div.booksContaining > ul > li:nth-child(1) '
+                                                     '> div > a > b')
```
m1yag1: In `pages/content_page.py` there is a Region called Book that represents a book in the book list. There is also a property on the page called `ContentPage.books_containing.books_list` where these elements should probably live.

m1yag1:
What issues does this PR link to?

BrianYing:
#227

PR #259, opened 2018-07-30
Added test case for empty collection publish not allowed

BrianYing:
added test case for empty collection publish not allowed, added selector for non-authorized user

### DIFF
--- a/pages/legacy/content_publish.py
+++ b/pages/legacy/content_publish.py
@@ -41,3 +41,7 @@ class ContentPublish(PrivatePage):
     @property
     def block_msg(self):
         return self.find_element(*self._block_msg_locator)
+
+    @property
+    def is_block_msg_displayed(self):
+        return self.block_msg.is_displayed()
--- a/pages/legacy/content_publish.py
+++ b/pages/legacy/content_publish.py
@@ -10,7 +10,9 @@ from selenium.webdriver.common.by import By
 class ContentPublish(PrivatePage):
     _publish_form_locator = (By.CSS_SELECTOR, '#content div.documentContent div div form')
     _message_textarea_locator = (By.CSS_SELECTOR, 'textarea[name="message"]')
-    _submit_button_locator = (By.CSS_SELECTOR, 'input[type="submit"][name="form.button.publish"]')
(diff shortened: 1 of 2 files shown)


pages/legacy/content_publish.py
```diff
@@ -10,7 +10,9 @@
 class ContentPublish(PrivatePage):
     _publish_form_locator = (By.CSS_SELECTOR, '#content div.documentContent div div form')
     _message_textarea_locator = (By.CSS_SELECTOR, 'textarea[name="message"]')
-    _submit_button_locator = (By.CSS_SELECTOR, 'input[type="submit"][name="form.button.publish"]')
+    _submit_button_locator = (By.CSS_SELECTOR, 'input[type="submit"][name="form.button.publish"], '
+                                               'input[type="submit"][name="form.button.submit"]')
+    _block_msg_locator = (By.CSS_SELECTOR, '#region-content > div > div > div > div > div')
```
Dantemss: :( was there any way to get a less flaky selector using classes maybe (it's fine if this is the only way).

tests/legacy/ui/test_create_import_publish_module_and_collection.py
```diff
@@ -170,3 +170,22 @@ def test_publish_collection(self, archive_base_url, legacy_base_url,
...
+
+        # WHEN the user clicks to create a new collection,
+        # agrees to the license and fills in the Title
+        cc_license = my_cnx.create_collection()
+        metadata_edit = cc_license.agree().submit()
+        collection_edit = metadata_edit.fill_in_title('CNX Automation Test Collection').submit()
+
+        # THEN click publish and publish is blocked (block msg, no submit button)
+        content_published = collection_edit.publish()
+        assert content_published.block_msg
```
Dantemss: I would say either assert that the block_msg is displayed (make a method that uses `is_element_displayed()`) or assert that it contains some text.

pages/legacy/content_publish.py
BrianYing: I am afraid that's the only way :(

BrianYing:
These failures are from Webview, unrelated to legacy. @Dantemss @m1yag1 @kerwinso

PR #268, opened 2018-08-03
Add test case for reorder-author

BrianYing:
also construct page object for role_edit

### DIFF
--- a/pages/legacy/collection_edit.py
+++ b/pages/legacy/collection_edit.py
@@ -17,6 +17,7 @@ class CollectionEdit(PrivatePage):
     _root_collection_locator = (By.CSS_SELECTOR, 'ul.x-tree-root-ct li.x-tree-node')
     _publish_link_locator = (By.CSS_SELECTOR, 'a[href$="collection_publish"]')
     _modal_locator = (By.CSS_SELECTOR, 'div.x-window')
+    _roles_tab_locator = (By.ID, 'contentview-roles')
 
     @property
     def username(self):
@@ -53,3 +54,9 @@ class CollectionEdit(PrivatePage):
         from pages.legacy.content_publish import ContentPublish
         content_publish = ContentPublish(self.driver, self.base_url, self.timeout)
         return content_publish.wait_for_page_to_load()
+
+    def roles(self):
+        self.find_element(*self._roles_tab_locator).click()
+        from pages.legacy.roles_edit import RolesEdit
+        roles_edit = RolesEdit(self.driver, self.base_url, self.timeout)
+        return roles_edit.wait_for_page_to_load()
--- a/pages/legacy/collection_edit.py
+++ b/pages/legacy/collection_edit.py
@@ -55,7 +55,7 @@ class CollectionEdit(PrivatePage):
         content_publish = ContentPublish(self.driver, self.base_url, self.timeout)
         return content_publish.wait_for_page_to_load()
 
-    def roles(self):
+    def click_roles_tab(self):
         self.find_element(*self._roles_tab_locator).click()
         from pages.legacy.roles_edit import RolesEdit
         roles_edit = RolesEdit(self.driver, self.base_url, self.timeout)
--- a/pages/legacy/collections_confirm_remove.py
+++ b/pages/legacy/collections_confirm_remove.py
@@ -2,23 +2,10 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from pages.legacy.base import PrivatePage
-
-from selenium.webdriver.common.by import By
+from pages.legacy.confirm_remove import ConfirmRemove
 
 
-class CollectionsConfirmRemove(PrivatePage):
-
-    _form_locator = (By.CSS_SELECTOR, 'form[action="folder_delete_confirm"]')
-    _remove_button_locator = (By.CSS_SELECTOR, 'input[type="submit"][name="form.button.delete"]')
-
-    @property
-    def form(self):
-        return self.find_element(*self._form_locator)
-
-    @property
-    def remove_button(self):
-        return self.form.find_element(*self._remove_button_locator)
+class CollectionsConfirmRemove(ConfirmRemove):
 
     def confirm(self):
         self.remove_button.click()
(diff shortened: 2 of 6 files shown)


pages/legacy/collections_confirm_remove.py
```diff
@@ -0,0 +1,27 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from pages.legacy.base import PrivatePage
+
+from selenium.webdriver.common.by import By
+
+
+class CollectionsConfirmRemove(PrivatePage):
+
```
m1yag1: Space can be removed

pages/legacy/roles_edit.py
```diff
@@ -0,0 +1,66 @@
...
+        def order_control(self):
+            return self.Order(self, self.find_element(*self._order_control_locator))
+
+        class Order(Region):
+            _top_locator = (By.CSS_SELECTOR, 'a:nth-child(1)')
+            _move_item_up_locator = (By.CSS_SELECTOR, 'a:nth-child(2)')
+            _move_item_down_locator = (By.CSS_SELECTOR, 'a:nth-child(3)')
+            _bottom_locator = (By.CSS_SELECTOR, 'a:nth-child(4)')
+
+            def top(self):
```
m1yag1: typically click events have that in the name... rename the methods to `click_top()`, `click_move_item_up()`, etc

pages/legacy/workspace_collection.py
```diff
@@ -0,0 +1,56 @@
...
+
+    @property
+    def select_all_checkbox(self):
+        return self.find_element(*self._select_all_checkbox_locator)
+
+    @property
+    def remove_button(self):
+        return self.find_element(*self._remove_button_locator)
+
+    def select_all(self):
```
m1yag1: rename to `click_select_all_checkbox`

pages/legacy/workspace_collection.py
```diff
@@ -0,0 +1,56 @@
...
+        self.select_all_checkbox.click()
+        return self
+
+    def remove(self):
```
m1yag1: rename to click_remove_button

pages/legacy/workspace_collection.py
```diff
@@ -0,0 +1,56 @@
...
+    @property
+    def collection_list(self):
+        return [self.WorkspaceCollectionEdit(self, el) for el in
+                self.find_elements(*self._collection_locator)]
+
+    class WorkspaceCollectionEdit(Region):
+        _link_locator = (By.CSS_SELECTOR, '.visualIcon > a')
+        _id_locator = (By.CSS_SELECTOR, 'td:nth-child(3)')
+
+        @property
```
m1yag1: click events are usually methods and not decorated with the property decorator. this way parens are required when calling the method like this:
`click_collection_link()`

pages/legacy/my_cnx.py
```diff
@@ -44,3 +50,9 @@ def workspace(self):
         from pages.legacy.workspace import Workspace
         workspace = Workspace(self.driver, self.base_url, self.timeout)
         return workspace.wait_for_page_to_load()
+
+    def workspace_collection(self):
```
m1yag1: rename to `click_workspace_collection`

pages/legacy/collection_edit.py
```diff
@@ -53,3 +54,9 @@ def publish(self):
         from pages.legacy.content_publish import ContentPublish
         content_publish = ContentPublish(self.driver, self.base_url, self.timeout)
         return content_publish.wait_for_page_to_load()
+
+    def roles(self):
```
m1yag1: rename to `click_roles_tab`

pages/legacy/collections_confirm_remove.py
```diff
@@ -0,0 +1,27 @@
...
+    @property
+    def form(self):
+        return self.find_element(*self._form_locator)
+
+    @property
+    def remove_button(self):
+        return self.form.find_element(*self._remove_button_locator)
+
+    def confirm(self):
```
m1yag1: Should confirm be clicking the remove button?

tests/legacy/ui/test_create_import_publish_module_and_collection.py
```diff
@@ -170,3 +170,41 @@ def test_publish_collection(self, archive_base_url, legacy_base_url,
...
+        test_col = -1
+        for col_idx in range(len(cols)-1):
+            if cols[col_idx].collection_id == collection_id:
+                test_col = col_idx
+
+        if test_col == -1:
+            pytest.skip('This test requires a specific CNX collection '
+                        'and test failed to find')
+
+        workspace_collection_edit = cols[test_col].click_collection_link
```
m1yag1: This should be a method call. you probably need to remove the `@property` decorator.

`click_collection_link()`

pages/legacy/workspace_collection.py
BrianYing: I am returning a specific type object. So I think the property decorator here is needed.

pages/legacy/collections_confirm_remove.py
BrianYing: not really sure. I think @Dantemss wrote this part and I modified it a bit for my own object. They share some similar features

pages/legacy/workspace_collection.py
```diff
@@ -0,0 +1,37 @@
...
+from regions.webview.base import Region
+from pages.legacy.workspace import Workspace
+
+from selenium.webdriver.common.by import By
+
+
+class WorkspaceCollection(Workspace):
+    _collection_locator = (By.CSS_SELECTOR, 'tr.odd, tr.even')
+
+    def remove(self):
```
m1yag1: Is this method used anywhere?

pages/legacy/roles_edit.py
```diff
@@ -0,0 +1,66 @@
...
+
+        @property
+        def order_control(self):
+            return self.Order(self, self.find_element(*self._order_control_locator))
+
+        class Order(Region):
+            _top_locator = (By.CSS_SELECTOR, 'a:nth-child(1)')
+            _move_item_up_locator = (By.CSS_SELECTOR, 'a:nth-child(2)')
+            _move_item_down_locator = (By.CSS_SELECTOR, 'a:nth-child(3)')
+            _bottom_locator = (By.CSS_SELECTOR, 'a:nth-child(4)')
```
Dantemss: Again would be nice if there were less flaky selectors for this, but if this is the only way, it's fine.

pages/legacy/workspace_collection.py
BrianYing: Not really. I created this object using the same structure as the Workspace since they similar. So I just built this method together which might be used in other test cases

PR #271, opened 2018-08-03
Added test for mathml3 valid in legacy

BrianYing:
Parametrize not done yet

### DIFF
--- a/pages/legacy/module_edit.py
+++ b/pages/legacy/module_edit.py
@@ -6,8 +6,6 @@ import re
 import xml.etree.ElementTree as ET
 
 from selenium.webdriver.common.by import By
-from selenium.webdriver.support.ui import WebDriverWait
-from selenium.webdriver.support import expected_conditions as EC
 
 from pages.legacy.base import PrivatePage
 from regions.legacy.base import Region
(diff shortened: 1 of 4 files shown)


BrianYing:
These failures are from Webview, unrelated to legacy. @Dantemss @m1yag1 @kerwinso

pages/legacy/module_edit.py
```diff
@@ -39,6 +45,14 @@ def title_header(self):
     def title(self):
         return self._title_regex.match(self.title_header.text).group(1)
 
+    @property
+    def portal_msg(self):
+        msg = WebDriverWait(self.driver, self.timeout).until(
+            EC.presence_of_element_located((By.CLASS_NAME, 'portalMessage'))
+        )
```
Dantemss: I would say instead of using the low-level `WebDriverWait` class you should define a method that uses `self.is_element_present()` to check for the `portalMessage`, then use this form: `self.wait.until(lambda _: self.is_portal_message_present)`

tests/legacy/ui/test_create_import_publish_module_and_collection.py
```diff
@@ -170,3 +170,30 @@ def test_publish_collection(self, archive_base_url, legacy_base_url,
...
+        module_edit = metadata_edit.fill_in_title('CNX Automation Test MathMl3').submit()
+
+        # AND click full source editing to edit
+        module_edit.edit_method.full_source_editing()
+        module_edit.edit_content_text(sample)
+
+        # THEN save and msg shows 'Saved'
+        module_edit.save()
+        # from time import sleep
+        # sleep(10)
```
Dantemss: Remove this commented out sleep :)

PR #334, opened 2018-09-12
Testing webview and archive integration using the latest images

pumazi:
This uses the latest webview and cnx-archive images to test a (partial) containerized environment.

Usage at the moment is `make test-webview`, which build the local selenium container, acquires the slim dump for data and executes the tests when things are ready.

This is a bit heavy... Lift with your legs not with your back. 😛

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,8 @@ $(STATEDIR)/docker-build: Dockerfile requirements.txt
 	mkdir -p $(STATEDIR)
 	touch $(STATEDIR)/docker-build
 
-# cnx_slim_dump.sql.gz:
-# 	scp backup2.cnx.org:`ssh backup2.cnx.org 'ls -t /var/backups/cnx_slim_dump.*.sql.gz | awk "{print $1; exit}"'` cnx_slim_dump.sql.gz
+cnx_slim_dump.sql.gz:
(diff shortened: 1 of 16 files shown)


Makefile
```diff
@@ -16,11 +31,18 @@ clean-pyc: ## remove Python file artifacts
...
 
+clean-state: ## remove the build state
+	rm -rf $(STATEDIR)
+
 clean-test: ## remove test and coverage artifacts
 	rm -fr .tox/
 	rm -fr assets/
 	rm -f report.html
 
+test: cnx_slim_dump.sql.gz $(STATEDIR)/docker-build
```
m1yag1: Would it be a terrible idea to add test commands for each of the markers? Maybe this can be handled via environment variable? Something like:
```
test_webview: cnx_slim_dump.sql.gz $(STATEDIR)/docker-build
	docker-compose up -d selenium-chrome
	docker-compose exec selenium-chrome wait-for db:5432 -- tox -- -m webview --webview_base_url=http://ui:8000 --archive_base_url=http://archive:6543
```

tests/webview/ui/test_content.py
```diff
@@ -774,7 +785,7 @@ def test_chapter_review_version_matches_book_version(webview_base_url, selenium,
 @markers.webview
 @markers.test_case('C195064')
 @markers.nondestructive
-@markers.parametrize('ch_review_id', ['4fGVMb7P@1'])
+@markers.parametrize('ch_review_id', ['e5fbbjPE@6'])
```
m1yag1: Perhaps we should leave off the @6 and always look at latest?

PR #343, opened 2018-09-24
Test the subject counts give an almost accurate response

pumazi:
We can't reliably test the counts for each subject when using the test dataset
(i.e. lean dump). When working with a test dataset
not all subjects will have content. Therefore, we test that
at least one of the subjects has a count. This effectively tests
the communication between the browser and the web api as well as the
browser's ability to render the results.

### DIFF
--- a/tests/webview/ui/test_browse.py
+++ b/tests/webview/ui/test_browse.py
@@ -46,13 +46,25 @@ def test_subject_categories_have_page_and_book_counts(webview_base_url, selenium
     # GIVEN the home page
     home = Home(selenium, webview_base_url).open()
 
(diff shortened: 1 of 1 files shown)


tests/webview/ui/test_browse.py
```diff
@@ -49,10 +49,22 @@ def test_subject_categories_have_page_and_book_counts(webview_base_url, selenium
     # When the browse link in the navbar is clicked
     browse = home.header.click_search()
 
-    # Then the subject categories have page and book counts
```
m1yag1: Please keep the `THEN` comment. 

https://trello.com/c/MljGhE2X/24-test-design-given-when-then

PR #344, opened 2018-09-24
Correct unicode title test expectations

pumazi:
The previous test for the figure's text isn't valid for this content, since
the figure itself doesn't have text. We'll end up with the same expected
results if we test for figures in general.

### DIFF
--- a/tests/webview/ui/test_content.py
+++ b/tests/webview/ui/test_content.py
@@ -410,14 +410,14 @@ def test_page_with_unicode_characters_in_title_loads(webview_base_url, selenium,
     # WHEN we load the content page
     content = content.open()
 
-    # Find a figure element
-    figure = content.content_region.figures[0]
+    # Ensure it has a figure element
+    assert content.content_region.has_figures
 
     # THEN the page does not reload afterwards
     # Wait 10 seconds to see if the page reloads
     sleep(10)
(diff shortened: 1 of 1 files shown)


tests/webview/ui/test_content.py
```diff
@@ -417,7 +417,7 @@ def test_page_with_unicode_characters_in_title_loads(webview_base_url, selenium,
     # Wait 10 seconds to see if the page reloads
     sleep(10)
     # If we don't get a StaleElementReferenceException then the page didn't reload
-    assert figure.text
+    assert figure.find_element_by_css_selector('img').get_attribute('alt')
```
m1yag1: There should not be any driver code in the test. The best thing to do here is to create a property method on the page object that returns True based on the locator you have here. As per the README.md:

> If you have WebDriver APIs in your test methods, You're Doing It Wrong.

The right thing to do is to create a locator on the page object and create a property method that accesses it, returns true, or whatever the best course of action is. There may be one you could already leverage on the existing page object.

pumazi:
Looks like this content isn't in production? 🤔

m1yag1:
what do you mean it isn't on production?

pumazi:
Correction, I thought we were running tests against production, but it's against qa.cnx.org

PR #376, opened 2018-10-16
Release the latest container build when tests pass for the master branch

pumazi:
This sets up the jenkins job to test branches as well as `master`. When `master` passes tests, it is released to the Container Registry as `openstax/cnx-automation:latest`, which then means it active for use by other `cnx-*` repos during their functional testing stages.

Ah yes... 🍹 development workflow simplified. 🏖

### DIFF
--- a/.jenkins/testing
+++ b/.jenkins/testing
@@ -1,15 +1,44 @@
 pipeline {
   agent { label 'docker' }
+  environment {
+    TESTING_CONTAINER_NAME = "cnx-func-testing-${env.BUILD_ID}"
+  }
   stages {
-    stage('Build') {
+    stage('Build Dev Container') {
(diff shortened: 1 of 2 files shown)


.jenkins/testing
```diff
@@ -1,17 +1,46 @@
 pipeline {
   agent { label 'docker' }
+  environment {
+    TESTING_CONTAINER_NAME = "cnx-func-testing-${env.BUILD_ID}"
```
pumazi: `BUILD_ID` isn't going to cut it, but I'll touch this up in a followup PR. The `BUILD_ID` is just a number, so it's possible that things could eventually collide, since the numbers are sequential by branch and pr, all starting at `1`.

PR #381, opened 2018-10-18
Remove the idea of locally testing the entire stack

pumazi:
It's likely not feasible to test the entire stack on a developer's machine. Some machines may be able to handle it, but most won't be able to. The amount of services that go into any of applications usually goes beyond what a developer of these tests will be able to obtain.

For now let's remove this notion and depend on a deployed stack. We can concentrate on running these tests against that stack rather than thinking about both how to run it on a local machine and in a deployed settings.

One of the reasons to just focus on a deployed setting is because CNX is very domain name oriented. It likes to have a domain name, which is something we can't rightly give it in a local setting.


pumazi:
BTW, I do have a copy of this in another branch as `docker-stack.staging.yml`, which is what I'm currently using for deploying this stack of services.

PR #464, opened 2019-01-03
Fixed test_navigation by checking for introduction first

m1yag1:
* Changed the beginning of the test to look for introduction rather than section number 1.
* Updated the section title so it doesn't strip out the section number
* Updated chapter_section_span locator to the new locator of `os-number`

### DIFF
--- a/regions/webview/content_item.py
+++ b/regions/webview/content_item.py
@@ -8,7 +8,7 @@ from regions.webview.base import Region
 
 
 class ContentItem(Region):
-    _chapter_section_span_locator = (By.CSS_SELECTOR, "span.chapter-number")
+    _chapter_section_span_locator = (By.CSS_SELECTOR, "span.os-number")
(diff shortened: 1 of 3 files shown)


m1yag1:
image
The test_navigation now passes due to the changes as indicated by the raw log in Travis

regions/webview/content_item.py
```diff
@@ -8,7 +8,7 @@
 
 
 class ContentItem(Region):
-    _chapter_section_span_locator = (By.CSS_SELECTOR, "span.chapter-number")
+    _chapter_section_span_locator = (By.CSS_SELECTOR, "span.os-number")
```
pumazi: Just curious... What does `os` mean here? I understood what `chapter` indicated, but I'm at a loss with `os`.

m1yag1: Not sure. They decided to change it to os-number. i'm assuming "os" is OpenStax. Maybe they wanted it to be more generic in the recipes?

PR #467, opened 2019-01-03
fixed test_in_book_search by included chapter section in title

m1yag1:
* The section title was not included from the search result. A property
was added that has the title and chapter section for the assert.


m1yag1:
test_in_book_search now passes per travis raw log. 

image

PR #478, opened 2019-01-08
Updated pytest, updated pytest-selenium, and fixed usage of global pytest.config

m1yag1:
* Updated pytest to 4.1.0
* Updated pytest-selenium to 1.15.1
* Added pytest_collection_modify_items function to conftest.py to mark
tests as run slow and replaced the marker from tests/markers.py to use
this new marker.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -99,6 +99,15 @@ def pytest_addoption(parser):
     )
 
 
+def pytest_collection_modifyitems(config, items):
+    if config.getoption("--runslow"):
+        return
+    skip_slow = pytest.mark.skip(reason="need --runslow option to run")
+    for item in items:
+        if "slow" in item.keywords:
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -99,6 +99,15 @@ def pytest_addoption(parser):
     )
 
 
+def pytest_collection_modifyitems(config, items):
```
pumazi: Does this get hooked into things by virtue of the `pytest_` prefix?

m1yag1: yes, during runtime it will collect all the items and mark them.

m1yag1: I basically did what they had in the docs ;).
https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option

PR #480, opened 2019-01-08
marking publishing test as xfail due to possible bug

m1yag1:
Marking legacy publishing test as xfail due to possible bug.

https://github.com/openstax/cnx/issues/220


m1yag1:
This doesn't seem to be an issue anymore. The test is passing as is now on master.

PR #540, opened 2019-02-28
initial commit for new visual test files

omehes:


tests/webview/visual/cnx_homepage_books.py
```diff
@@ -0,0 +1,480 @@
...
+import os
+from tests.utils import cnx_deploy_version
+
+load_dotenv(verbose = True)
+
+applitools_key = os.environ['APPLITOOLS_API_KEY']
+url_addr = os.environ['URL_HOME_PAGE']
+App_Name = 'CNX HOME PAGE BOOK-ICONS'
+
+class Books_Home_Page_Layout:
```
m1yag1: This isn't in proper class format. When we name classes in Python there are no underscores. This should be named

`class BooksHomePageLayout:`

Additionally, if you want to group your tests within a Class you should implement the tests as methods like so.

```python
class MyAwesomeTest(object):
    def test_my_awesome_page_layout(self):
        print("everything is awesome")
```
You need to refactor this class to look something like this.

m1yag1: To further build on this concept we can pass into our tests fixtures that are available to all the other tests. for example.

```python
class MyAwesomeTest(object):
    def test_my_awesome_page_layout(self, selenium, webview_base_url, request):
        page = selenium.get(webview_base_url)        
        
        eyes = Eyes(request.config.applitools_api_key)

      . . .
```

tests/webview/visual/cnx_homepage_books.py
```diff
@@ -0,0 +1,480 @@
...
+
+    # current cnx deploy and webview versions
+    cnx_ver = cnx_deploy_version.cnx_versions
+    cnx_web_web = cnx_deploy_version.cnx_w_web
+
+
+    try:
+
+        # Open a Chrome browser
+        driver = webdriver.Chrome()
```
m1yag1: See my comment about using fixtures: https://github.com/openstax/cnx-automation/pull/540/files#r261679274

You don't need to implement driver again here because it's already provided by the pytest-selenium plugin. You simply need to refactor this like mentioned before and you will have the driver available.

tests/webview/visual/cnx_homepage_books.py
m1yag1: Reference in legacy tests for how to organize as a class:

https://github.com/openstax/cnx-automation/blob/master/tests/legacy/ui/test_create_import_publish_module_and_collection.py#L20-L43

tests/webview/visual/test_homepage_collections.py
```diff
@@ -0,0 +1,422 @@
...
+        # Visual checkpoint - screenshot #1
+        applitools.check_window('{}'.format(deploy_version_tag))
+
+        # finds element by xpath, scrolls it into view and clicks "Show more" links
+        element = selenium.find_element_by_xpath("//*[@id='featured-books']/div/div[1]/div/div[1]/a/img")
+        selenium.execute_script("return arguments[0].scrollIntoView(true);", element)
+
+        sleep(1)
+
+        element2 = selenium.find_element_by_xpath("//*[@id='featured-books']/div/div[1]/div/div[1]/div/a[1]")
```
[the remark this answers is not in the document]
m1yag1: Do a thought experiment and we can meet in order to think it through together. What if you had something that exposed a list of every item. something like this:

```
for book in home.featured_books:
    if book.is_displayed
    ...
```

You don't necessarily have to iterate over it like that.. Could you chunk it? Could you iterate over it until you found an item that was not displayed on the page letting you know it was off the page?

fixtures/webversions.py
```diff
@@ -0,0 +1,48 @@
+from dotenv import load_dotenv
+import os
+import pytest
+import urllib.request
+from ast import literal_eval
+load_dotenv(verbose = True)
+url_addr_version = os.environ['URL_ADDR_VERSION']
```
m1yag1: You are using `URL_ADDR VERSION` here... environment variables should be handled somewhere else. What could you do differently?

fixtures/webversions.py
```diff
@@ -0,0 +1,48 @@
...
+
+# adding cnx-deploy version number to applitools screenshot results
+@pytest.fixture(scope="session")
+def deploy_version_tag(request):
+    """Returns cnx version numbers"""
+
+    version_page = urllib.request.urlopen(url_addr_version)
```
m1yag1: `version_page = urllib.request.urlopen(url_addr_version)`

We should try to avoid globals. What's another way for you to pass `url_addr_version` into this function?

fixtures/webversions.py
```diff
@@ -0,0 +1,48 @@
...
+        raise Exception("cnx deploy version is missing")
+    else:
+        yield cnx_deploy
+
+# adding webview version number to applitools screenshot results
+@pytest.fixture(scope="session")
+def webview_version_tag(request):
+    """Returns cnx version numbers"""
+
+    version_page = urllib.request.urlopen(url_addr_version)
```
m1yag1: You have basically the same code in two places. This is repetitive and we don't like to repeat the same code in different places. What could you do to DRY (Don't Repeat Yourself)?

tests/webview/visual/test_homepage_collections.py
```diff
@@ -0,0 +1,422 @@
+from time import sleep
+from applitools import MatchLevel
+from tests import markers
+
+from pages.webview.home import Home
+
+APP_NAME = 'CNX HOMEPAGE BOOKS3'
```
m1yag1: is this the best name? What does this mean?

tests/webview/visual/test_home_objects.py
```diff
@@ -0,0 +1,36 @@
...
+@markers.webview
+@markers.nondestructive
+def test_home_objects(applitools, webview_base_url, selenium):
+
+    try:
+
+        home = Home(selenium, webview_base_url).open()
+
+        # Start the test and set the browser's viewport size to 1400x820
+        applitools.open(driver=selenium, app_name=APP_NAME, test_name='Test homepage books', viewport_size={'width': 1400, 'height': 820})
```
m1yag1: We should try to avoid magic numbers. We may want to pull this off of the home page. something like this:

`applitools.open(driver=selenium, app_name=APP_NAME, test_name='Test homepage books', viewport_size={'width': home.window_width, 'height': home.window_height})`

tests/webview/visual/test_home_objects.py
```diff
@@ -0,0 +1,36 @@
+from tests import markers
+from pages.webview.home import Home
+
+APP_NAME = 'CNX HOMEPAGE BOOKS VISUAL VERIFICATION-TEST1'
```
m1yag1: We need to come up with a set way to describe the app. I think it's fine if we set this to `Webview` and leave it like that for when this goes into production.

tests/webview/visual/test_home_objects.py
```diff
@@ -0,0 +1,36 @@
...
+
+        # End the test.
+        applitools.close()
+
+    finally:
+
+        # If the test was aborted before applitools.close was called, ends the test as aborted
+        applitools.abort_if_not_closed()
+
+        assert applitools.api_key
```
m1yag1: This needs to go away.

tests/webview/visual/test_home_objects.py
m1yag1: if home.window_width doesn't exist it will need to be implemented.

tests/webview/visual/test_home_objects.py
omehes: Gone! I forgot to remove after our previous conversation.

tests/webview/visual/test_homepage_collections.py
omehes: I changed it to webview_homepage. In my other test case I work on, I would name it webview_pages as it will check pages

tests/webview/visual/test_home_objects.py
omehes: I am using this logic adopted from test_home.py:
```
@markers.parametrize("width,height", [(1400, 820)])
def test_home_objects(applitools, webview_base_url, selenium, width, height):
...
    try:
        home = Home(selenium, webview_base_url).open()

        # Start the test and set the browser's viewport size to 1400x820
        applitools.open(driver=selenium, app_name=APP_NAME, test_name='Test homepage books', viewport_size={'width': width, 'height': height})
```
This seems to be working and takes away magic numbers.

PR #576, opened 2019-05-28
Create integration tests for rex redirects

BryanEliDimas:
**Solves issue**: https://github.com/openstax/cnx/issues/358


pumazi:
@mike, Everything is ready to roll with regards to these tests, unless you spot something funny. The only thing to mention is that they are and will be failing up until we result REX (Aug?). Thoughts?

BryanEliDimas:
Merging myself because @m1yag1 is on vacation and it's code reviewed ✅

PR #582, opened 2019-06-03
Update README.md

BryanEliDimas:
Resolves various issues I came across and opened recently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,9 @@
 <a href="https://github.com/ambv/black"><img alt="Code style: black" src="https://img.shields.io/badge/code%20style-black-000000.svg"></a>
 </p>
 
-## Getting started
+# Getting started
+
+There are two ways to run the tests, using Docker or with a local virtual environment. Not all tests require Selenium, but for the ones that do you will need `chromedriver` locally installed (in MacOS with homebrew run `brew cask install chromedriver`).
 
 ### Clone the repository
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,7 +5,9 @@
 <a href="https://github.com/ambv/black"><img alt="Code style: black" src="https://img.shields.io/badge/code%20style-black-000000.svg"></a>
 </p>
 
-## Getting started
+# Getting started
+
+There are two ways to run the tests, using Docker or with a local virtual environment. Not all tests require Selenium, but for the ones that do you will need `chromedriver` locally installed (in MacOS with homebrew run `brew cask install chromedriver`).
```
pumazi: I could be wrong, but I believe we want to enforce the notion of there only being one way to run this. That helps mitigate the issues around documenting setup for all platforms, which turns into a tangled mess of unmaintainable documentation. (see cnx-archive for example 😬)

BryanEliDimas: Hm... so then we should delete the part of the README for "preparing the project locally" then?

BryanEliDimas:
@pumazi The legacy and webview tests are giving a weird message:
```
{'message': 'session not created: This version of ChromeDriver only supports '
            'Chrome version 74\n'
            '  (Driver info: chromedriver=74.0.3729.169 '
```

README.md
BryanEliDimas: So then should we only make mention of Docker and omit local installation?

PR #585, opened 2019-06-12
chore: clarify CLI args

philschatz:
Adds more description to the CLI args to show which ones are used by Travis and how others might be useful.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -34,25 +34,25 @@ def pytest_addoption(parser):
         "--disable-dev-shm-usage",
         action="store_true",
         default=os.getenv("DISABLE_DEV_SHM_USAGE", False),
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -63,7 +63,7 @@ def pytest_addoption(parser):
         "--runslow",
         action="store_true",
         default=os.getenv("RUNSLOW", False),
-        help="run slow tests.",
+        help="run slow tests (legacy).",
```
m1yag1: It's not just for legacy. There is a test that checks 80 or so pages for `not converted yet` messages in the American Government book.

m1yag1: perhaps `(mostly legacy)`

PR #586, opened 2019-06-18
Json search

omehes:


m1yag1:
Hey @omehes , cnx-automation is probably not the right place for this script. cnx-automation is a testing framework that utilizes pytest. Unless your script can be written so pytest runs the code as part of a test and does assertions I don't think this is the right place.

What i would recommend is that you create a repo for these kinds of scripts. I could also create it for you. You could call it `ce-qa-tools` or something like that. Let me know if you want me to create that.

PR #589, opened 2019-06-26
Check not self closing

therealmarv:
add testcases for not self closing tags with verification code


tests/legacy/ui/test_no_selfclose_create_import_publish_module_and_collection.py
```diff
@@ -0,0 +1,119 @@
...
+        my_cnx = login_page.login(legacy_username, legacy_password)
+
+        # WHEN the user clicks to create a new module, agrees to the license and fills in the Title
+        cc_license = my_cnx.create_module()
+        metadata_edit = cc_license.agree().submit()
+        module_edit = metadata_edit.fill_in_title("CNX Automation Test Module").submit()
+
+        # THEN the user is brought to the module editor
+        assert type(module_edit) is ModuleEdit
+        assert module_edit.title == "CNX Automation Test Module"
```
m1yag1: Might want to title it in regards to self closing tags so we can differentiate from other tests.

therealmarv: changed on https://github.com/openstax/cnx-automation/pull/589/commits/2885e2f2c2c6fca3fcbe25d72696853524516e46

therealmarv:
The JSON snapshot here https://github.com/openstax/cnx-automation/blob/1f9e51071082ef284471822e95818b872b3305cd/snapshots/json/legacy/not_self_closing.json is with self closing tags.

It should be once if qa is updated
1. deleted
2. regenerated
3. verified by human if `<!-- no-selfclose -->` is on all important places present.

tests/legacy/ui/test_no_selfclose_create_import_publish_module_and_collection.py
```diff
@@ -0,0 +1,119 @@
...
+from pages.legacy.module_edit import ModuleEdit
+from pages.legacy.collection_edit import CollectionEdit
+from pages.legacy.content_published import ContentPublished
+from pages.archive.legacy_content import LegacyContent
+
+from regions.legacy.collection import Collection
+from regions.legacy.module import Module
+
+
+class TestNoSelfCloseCreateImportPublishModuleAndCollection(object):
```
m1yag1: I don’t think you are publishing a module and a collection.

therealmarv:
This is ready for review, I will create another PR ("release early, release often") for collection publishing test.

m1yag1:
To go along with @karenc 's  comment, I was expecting the test to be expecting to fail. This would pass w/ the `pytest.xfail` marker. When QA was updated w/ the change the test to should PASS which pytest would raise as a failure. In that event, we would delete the snapshot and recreate to store as the baseline snapshot.

PR #594, opened 2019-07-22
Changes to cnx-automation test files: test_browse.py and test_content…

omehes:
Latest changes:

test_content.py
"C176231" - non-existent test case. Removed it from the @markers.test_case(...)
"C176260" - outdated (no google+ in cnx). Removed it from the @markers.test_case(...)
"C162195" - very old bug, not present in trello anymore. Removed it from @markers.test_case(...)
Also, it is verified in test case C132548

test_browse.py
"C176268" - removed this test case from this function's @markers.test_case(...)
and added to the one below - "def test_subject_categories_load(webview_base_url, selenium):"
"C176270" - this is a non-existent test case. Instead, C176268 test case verifies this.
Replacing "C176270" with C176268

….py, removing/changing test cases in markers in 
test_browse.py
test_content.py 

Added notes to both test cases to affected functions.

### DIFF
--- a/pages/webview/content.py
+++ b/pages/webview/content.py
@@ -281,6 +281,7 @@ class Content(Page):
             _root_locator = (By.CSS_SELECTOR, "#content div.pinnable div.share")
             _facebook_share_link_locator = (By.CSS_SELECTOR, "ul li a.facebook")
             _twitter_share_link_locator = (By.CSS_SELECTOR, "ul li a.twitter")
+            _linkedin_share_link_locator = (By.CSS_SELECTOR, "ul li a.linkedin")
 
             @property
             def is_facebook_share_link_displayed(self):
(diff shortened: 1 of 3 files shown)


tests/webview/ui/test_content.py
```diff
@@ -452,22 +455,22 @@ def test_get_this_book(webview_base_url, selenium):
     if offline_zip_displayed:
         assert downloads.is_offline_zip_available
 
+# !!! old bug verification from March 2017. Test case was selected to be deleted. We should remove the whole function
```
m1yag1: If it's old then remove it. We want to avoid keeping commented code as much as possible.

omehes: Removed the old commented out function and all the comments regarding changes.

pages/webview/content.py
```diff
@@ -306,6 +307,18 @@ def twitter_share_link(self):
             def twitter_share_url(self):
                 return self.twitter_share_link.get_attribute("href")
 
+            @property
+            def is_linkedin_share_link_displayed(self):
```
m1yag1: this should be `linkedin_share_link_is_displayed`

PR #600, opened 2019-08-08
Fixed logging into legacy when using Chrome 76

m1yag1:
For some reason Chrome 76 and the updated chromedriver was creating
errors when submitting the login form for legacy.

* Change action from form.submit() to actually clicking on the submit
button.


omehes:
I checked it and we also discussed it. Looks good. "selenium.webdriver.common.keys import Keys" will be removed?

PR #605, opened 2019-08-22
Enable xfail rex redirect tests

omehes:
for Karen to view and use - changes made to test 6 books on rex.


karenc:
I'm going to remove the extra per page rex redirect tests for now so that we can get the existing rex redirects tests running.

m1yag1:
Placing slack convo here for posterity

```
@karen. I have a few questions about the rex redirects in cnx-automation. Starting a thread here. :smiley:
:+1:
1


mike  6 hours ago
How come we aren’t using a fixture that reads in the redirects?

karen:palm_tree:  6 hours ago
I'm not sure... what do you mean?  I'm not very familiar with cnx-automation

mike  6 hours ago
I’ll add these comments to the PR btw. not trying to split the information two places just wanted more instant feedback hehe.

karen:palm_tree:  6 hours ago
Yeah, I was going to suggest that :stuck_out_tongue:

mike  6 hours ago
well.. the urls are hardcoded https://github.com/openstax/cnx-automation/pull/605/files#diff-df6d47ee71022eb28c67ae056e77b570R31

mike  6 hours ago
i tend to try and avoid hardcoded urls and pass them in via parametrize.

karen:palm_tree:  6 hours ago
That's a good idea

karen:palm_tree:  6 hours ago
I just left the test as it was

karen:palm_tree:  6 hours ago
Except the book it used wasn't working :stuck_out_tongue:

mike  6 hours ago
ah so according to git blame this was brian’s work. hehe

karen:palm_tree:  6 hours ago
Right :rolling_on_the_floor_laughing:

mike  6 hours ago
ok i’ll think on it a sec..

karen:palm_tree:  6 hours ago
Would you put the fixture in the test file or somewhere else?

mike  6 hours ago
depends on how many we would want to test.


karen:palm_tree:  6 hours ago
Actually do we already have the urls of books in fixtures?

mike  6 hours ago
for many of them they are with the test using parametrize.

mike  6 hours ago
example: https://github.com/openstax/cnx-automation/blob/master/tests/webview/ui/test_content.py#L118-L126

tests/webview/ui/test_content.py:118-126
@markers.parametrize(
    "id",
    [
        "AgQDEnLI@10.1:TrIRM88K@5",
        "AgQDEnLI@10.1:XZe6d2Jr@9",
 Show more
<https://github.com/openstax/cnx-automation|openstax/cnx-automation>openstax/cnx-automation | Added by GitHub

karen:palm_tree:  6 hours ago
Ah ok, that's simple
```

m1yag1:
I recommend running these tests locally.
If you are to run them normally (w/o -n) it will take over 23 min.

To run them faster you'll want to run them using multiple workers. I used 8 workers to get the time under 4 min w/ this command.

`pytest -n 8 -m rex -k test_chemistry_2e_urls_redirect_to_rex -vvv `

PR #611, opened 2019-09-18
added testrail test cases, Biology 2e, new marker

omehes:
4 files were modified:
- created a new test file: test_rex_redirects_testrail.py and
- added test_case markers to mark test results in testrail
- added new marker: mark.skip
- added new uuid.txt file for Biology 2e
- added biology_2e_uri function in webview.py


m1yag1:
Why the new file name `test_rex_redirects_testrail`? we shouldn't have to namespace files to specify tests because we have markers.

omehes:
@m1yag1 I just named it that way, so I know it is the testrail version. But if that is not needed, I can change it back to its original name.

tests/webview/integration/test_rex_redirects_testrail.py
```diff
@@ -0,0 +1,156 @@
+import backoff
```
m1yag1: Consolidate the files. There doesn't need to be a testrail specific file. if you need to see what tests are running you can always use --collect-only in the pytest command. There is a lot of duplication with the two files that isn't necessary

omehes: OK. I rename the file back to its original name and re-submit the changes.

omehes: @m1yag1 pushed the file name change, corrected and added some url in biology.txt file and added mark.otto to markers.py to have it available in the future too

tests/markers.py
```diff
@@ -13,6 +13,8 @@
 parametrize = mark.parametrize
 test_case = pytestrail.case
 xfail = mark.xfail
+skip = mark.skip
+otto = mark.otto
```
m1yag1: Remove otto marker. If you need a marker as your working then add it on the fly but we shouldn't have user specific info with the framework. A person will come along later and ask, "What does the otto marker mean?" and no one will know.

omehes: @m1yag1 marker removed. pushed changes. the circleci check came up with 2 fails in webview tests. they are completely unrelated and passed previously.

omehes: @m1yag1 ``` assert False  +  where False = <pages.webview.base.Page.Footer object at 0x7f133da83be0>.is_android_app_link_displayed ```

android app link has been removed on staging. that is the fail

omehes: @m1yag1 
1. I removed android app link from test_home in another PR - this is OK
2. the other fail is related to the fact that the book redirects:
```
Failed for this reason:

test_in_book_search[e42bd376-624b-4c0f-972f-e0c57998e765-mitosis genetics-True-0-True-False] - tests.webview.ui.test_content

tests/webview/ui/test_content.py
webview_base_url = 'https://staging.cnx.org'
```
Book e42bd376-624b-4c0f-972f-e0c57998e765 redirects. Not related to this change - OK

tests/markers.py
```diff
@@ -13,6 +13,7 @@
 parametrize = mark.parametrize
 test_case = pytestrail.case
 xfail = mark.xfail
+skip = mark.skip
```
m1yag1: Where are you using the skip marker? I'm not seeing it on any of the tests.

omehes: I removed the function which used it. Removed the marker skip.

tests/webview/integration/test_rex_redirects.py
```diff
@@ -28,6 +28,8 @@ def test_redirect_for_rex_books(webview_base_url, rex_base_url):
     assert rex_base_url in response.history[0].headers["location"]
 
 
+@markers.test_case("C553086")
+@markers.slow
```
m1yag1: Soo, you have to put slow on all of these for them to be updated in testrail?

omehes: yes. otherwise no testrail marking

omehes: @m1yag1 removed some of the slow markers as per our discussion and pushed changes

omehes: both test fails are the same as previously explained (removed app link -> PR waiting to be approved and redirected book). not related.

PR #612, opened 2019-09-19
removed check for android app link from test_home.py

omehes:
removed the test for the android app link in footer. it is no longer present on cnx


omehes:
Failed for this reason: 
```test_in_book_search[e42bd376-624b-4c0f-972f-e0c57998e765-mitosis genetics-True-0-True-False] - tests.webview.ui.test_content
tests/webview/ui/test_content.py
webview_base_url = 'https://staging.cnx.org'```

Not related to this change.

PR #613, opened 2019-09-25
changed book uuid from Biology to Organizational Behavior

omehes:
Changed book uuid to a non-redirecting book, changed search term to a non-existing and has_results to False

("2d941ab9-ac5b-4eb8-b21c-965d36a4f296", "filanthropic", False, 0, True, False)

### DIFF
--- a/tests/webview/ui/test_content.py
+++ b/tests/webview/ui/test_content.py
@@ -320,7 +320,7 @@ def test_share_on_top_right_corner(webview_base_url, selenium):
             None,
             None,
         ),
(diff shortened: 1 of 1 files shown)


tests/webview/ui/test_content.py
```diff
@@ -320,7 +320,7 @@ def test_share_on_top_right_corner(webview_base_url, selenium):
             None,
             None,
         ),
-        ("e42bd376-624b-4c0f-972f-e0c57998e765", "mitosis genetics", True, 0, True, False),
```
m1yag1: I think we already have one above that doesn't return results. this test should be set to `True` and have a term that gets results.

PR #614, opened 2019-09-26
Checking book titles redirecting to rex

omehes:
Created a fixture in fixtures/webview.py, made changes to and removed some functions from test_rex_redirects.py and added redirected_books.txt containing book titles which redirects (needs to be updated)


omehes:
@m1yag1 Fail is due to the PR not yet approved and merged - search issue with redirecting book
Not related to this PR

tests/webview/integration/test_rex_redirects.py
```diff
@@ -90,36 +76,6 @@ def test_minimal_view_for_android_apps(webview_base_url, rex_base_url):
     assert rex_base_url not in response.url
 
 
-@markers.rex
-@markers.nondestructive
-@markers.xfail  # until we resolve and deploy https://github.com/openstax/cnx/issues/357
-def test_cnx_sitemap_exclusion(rex_base_url, archive_base_url):
```
m1yag1: Why are we removing this? This is a test set to xfail b/c it hasn't been implemented yet. When we do this work we'll want to check that the books that are on rex are not on the sitemap

fixtures/webview.py
```diff
@@ -85,3 +85,22 @@ def webview_instance(webview_base_url):
...
+
+
+@pytest.fixture
+def redirecting_books_titles():
+    """Returns all the books from the list of redirecting books to rex
+    """
+
+    here = os.path.join(DATA_DIR, "redirected_books.txt")
+
+    f = open(here, "r")
```
m1yag1: It's a best practice in python to use a context manager. there are other examples of this in the framework. You'll want to implement this like:

`with open(filename, "mode") as file:` ...

fixtures/webview.py
```diff
@@ -85,3 +85,22 @@ def webview_instance(webview_base_url):
...
     else:
         return "prod"
+
+
+@pytest.fixture
+def redirecting_books_titles():
+    """Returns all the books from the list of redirecting books to rex
+    """
+
+    here = os.path.join(DATA_DIR, "redirected_books.txt")
```
m1yag1: You don't need this. At the top of the file is a global called `DATA_DIR`. Please use other examples to your advantage.

tests/webview/integration/test_rex_redirects.py
omehes: I removed it due to the comment from Bey in the PR: 'Closing this card because the work no longer needs to happen due to other SEO work being completed.'

fixtures/webview.py
```diff
@@ -85,3 +85,20 @@ def webview_instance(webview_base_url):
...
+    """
+
+    here = DATA_DIR + "/redirected_books.txt"
+
+    with open(here, "r") as file:
+
+        tlines = file.readlines()
+        ttlines = map(lambda each: each.strip("\n"), tlines)
+
+        file.close()
```
m1yag1: You don't need to close when you use a context manager.

fixtures/webview.py
```diff
@@ -85,3 +85,20 @@ def webview_instance(webview_base_url):
...
     else:
         return "prod"
+
+
+@pytest.fixture
+def redirecting_books_titles():
+    """Returns all the books from the list of redirecting books to rex
+    """
+
+    here = DATA_DIR + "/redirected_books.txt"
```
m1yag1: Use `os.path.join(pathA, pathB)` to join file paths. Did you see my other comment about looking at other examples? There is already a fixture that gets you the list of books.

https://github.com/openstax/cnx-automation/blob/master/fixtures/webview.py#L70-L78

omehes:
@m1yag1 made changes: deleted my fixture and used the existing one, re-wrote the function, added 2 new books to check selection of pages, added uuid.txt files with links for the 10 redirecting books.

I forgot to update my branch, the search issue still fails... not related to my changes

tests/webview/integration/test_rex_redirects.py
```diff
@@ -31,28 +30,24 @@ def test_archive_is_still_reachable(archive_base_url, rex_base_url):
...
 
-    for eligible_book in redirecting_books_titles:
-        # GIVEN the home page
-        home = Home(selenium, webview_base_url).open()
+    for book in rex_released_books:
 
-        # WHEN we click on a REX book
-        for book in home.featured_books.openstax_list:
+        # GIVEN the cnx url of a redirecting book
+        cnx_book_url = f"{webview_base_url}/contents/{book}"
```
m1yag1: I'm not sure this is a good test. You are building the url yourself and then making a request. I believe the intent of the test is to click the link from the webview homepage no?

omehes:
@m1yag1 Sorry for this mess. I did the following:
- I made changes to test_content.py (new book and search string) because git rebase did not updated for some reason this branch
- then started to work on the 'test_redirecting_to_rex_from_within_webview' function as per your review
- then decided that at least the new books in new functions should be merged, so I reverted the above function to its master state and git push changes
- there was a conflict between the new book/search term. this I resolved and re-pushed
The major change is the new books/pages added (microbiology, concepts of biology and added more links to biology2e.txt

m1yag1:
There seems to be more test data files for redirects than there is fixtures/tests.

omehes:
@m1yag1 deleted 6 of 10 uuid.txt files not related to this PR

fixtures/data/webview/redirected_books.txt
```diff
@@ -0,0 +1,10 @@
+Chemistry 2e
```
m1yag1: Are you still using this file?

omehes: no. I remove it. I overlooked this...

omehes:
deleted the txt file

PR #615, opened 2019-09-30
tests rex redirection of all eligible books from webview to rex

omehes:
Amended the 'test_redirecting_to_rex_from_within_webview' function to test all the redirecting books by clicking on featured books in cnx (as per your review notes from another (merged) PR (https://github.com/openstax/cnx-automation/pull/614))


omehes:
@m1yag1 finally, ready for review :)

tests/webview/integration/test_rex_redirects.py
```diff
@@ -27,28 +27,36 @@ def test_archive_is_still_reachable(archive_base_url, rex_base_url):
...
-def test_redirecting_to_rex_from_within_webview(webview_base_url, rex_base_url, selenium):
-    """Webview needs to redirect to REX when one of the featured books is a REX book.
-    https://github.com/openstax/cnx/issues/401
-    """
+def test_redirecting_to_rex_from_within_webview(
+    webview_base_url, selenium, rex_base_url, rex_released_books
+):
+
+    # GIVEN the home page and REX released books list
+    for allbooks in rex_released_books:
```
m1yag1: usually the target in the for loop is singular. I suggest renaming this to `book` instead of `allbooks`.

For example:

```
for book in rex_released_books:
...
```

tests/webview/integration/test_rex_redirects.py
```diff
@@ -27,28 +27,36 @@ def test_archive_is_still_reachable(archive_base_url, rex_base_url):
...
+        for book in home.featured_books.openstax_list:
+
+            # WHEN we click on a featured book
+            if book.cnx_id in allbooks:
+
+                book_title = book.title
 
-    # GIVEN the home page
-    home = Home(selenium, webview_base_url).open()
+                link_elem = home.selenium.find_element_by_link_text(book_title)
```
m1yag1: Do not place selenium code in tests. In the [README.md](https://github.com/openstax/cnx-automation#brief-architectural-description) this is mentioned.

According to Siman Stewart,

> If you have WebDriver APIs in your test methods, You're Doing It Wrong.

tests/webview/integration/test_rex_redirects.py
```diff
@@ -27,28 +27,36 @@ def test_archive_is_still_reachable(archive_base_url, rex_base_url):
...
+    for allbooks in rex_released_books:
+
+        home = Home(selenium, webview_base_url).open()
+
+        for book in home.featured_books.openstax_list:
+
+            # WHEN we click on a featured book
+            if book.cnx_id in allbooks:
+
+                book_title = book.title
```
m1yag1: Is the `for allbooks in rex_released_books` even necessary? You are already using an if statement in the for loop. Could you use that better?

omehes: This is to prevent stale element exception by resetting to webpage as it changes during the test. If I remove that, I am getting that exception.

tests/webview/integration/test_rex_redirects.py
omehes: If I use Page object (click_book_cover), the redirected webpage get stuck as it expects to have a cnx page and not a rex page. I tried various things regarding this and this is what worked for me.

omehes:
@m1yag1 pushed the changes we talked about. 
First fail is due the fact that the new redirecting books are not yet on staging.cnx.org (like Calculus vol 1, etc)
Second fail is in text_content file. Not sure why it happens but not related to my changes.

Issue #549, opened 2019-03-08
test_chapter_review_version test fails due to change in baked book process of creating the chapter review section


m1yag1:
https://github.com/openstax/cnx-automation/blob/master/tests/webview/ui/test_content.py#L826-L837

This test is currently failing because the Chapter Review page is no longer a single page containing sections for key terms, key equations, etc. The Chapter Review has been turned into a sub-collection consisting of modules that were once the sections of the page.

This test needs to be updated to account for this change. This change will be in the next production release targeted for 3/12/2019.

PR #620, opened 2019-10-04
fixed failed functions in test_content due to redirecting books

omehes:
changed redirecting books to non-redirecting ones
and in 'test_chapter_review_version_matches_book_version' I removed the xfail (# marks=markers.xfail(reason="https://github.com/openstax/cnx-automation/issues/549"))


m1yag1:
@omehes is there a reason ce-scripts are merged in here?

PR #622, opened 2019-10-05
test_content: fix failing tests due to redirecting books

omehes:


omehes:
@m1yag1 Made the changes and removed 12 files not related. Now it should be only what I wanted to commit and push.

PR #626, opened 2019-10-21
removed visual tests - applitools - and related references, fixtures

omehes:
as per our discussion last week, I removed all applitools tests and related files, fixtures, etc


m1yag1:
The applitools dependencies need to be removed from the requirements.txt file. This is an interesting process in and of itself. What you'll need to `pip install pip-tools`. Remove applitools dependencies from requirements.txt and then use `pip-sync`. This is a good guide. We already use pip-tools btw. That's what the `requirements.in` file is for.

PR #627, opened 2019-10-23
Test for the inclusion and exclusion of derived copies in the sitemap

pumazi:
See openstax/cnx#727 for details

Note, the adjoining backend logic is yet to come...

### DIFF
--- a/tests/webview/ui/test_sitemap.py
+++ b/tests/webview/ui/test_sitemap.py
@@ -15,6 +15,7 @@ def xor(a, b):
 
 
 def xnor(a, b):
+    # See truth table: https://en.wikipedia.org/wiki/XNOR_gate
     return not xor(a, b)
 
 
--- a/tests/webview/ui/test_sitemap.py
+++ b/tests/webview/ui/test_sitemap.py
@@ -2,6 +2,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import requests
+import lxml.etree
+
 from tests import markers
 
 from pages.webview.sitemap_index import SitemapIndex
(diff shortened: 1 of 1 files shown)


tests/webview/ui/test_sitemap.py
```diff
@@ -31,3 +34,47 @@ def test_sitemap_is_segmented_by_author(webview_base_url, selenium, author_usern
...
+    # WHEN a spider requests the specific author's sitemap
+    response = requests.get(url)
+    response.raise_for_status()
+    xml = lxml.etree.fromstring(response.content)
+    #: give the text for the the element named 'loc' in any namespace
+    urls = xml.xpath("//*[local-name()='loc']/text()")
+
+    # THEN non-OpenStax derived copies have been removed
+    # while OpenStax books remain
+    presents_operation = be_present and (lambda x: x) or (lambda x: not x)
```
m1yag1: I'd break this out a bit more for readability, either a conditional or named function. It definitely shows advanced usage of python features like lambda but it does so at the expense of readability. Upon first looking at the code I had to scratch my head for a moment, mostly in thinking about what `be_present` meant. This is a common thing with lambdas in my experience; they are particularly prone to poor naming. Additionally, I'm concerned about Otto maintaining this

tests/webview/ui/test_sitemap.py
```diff
@@ -31,3 +42,47 @@ def test_sitemap_is_segmented_by_author(webview_base_url, selenium, author_usern
...
+    response = requests.get(url)
+    response.raise_for_status()
+    xml = lxml.etree.fromstring(response.content)
+    #: give the text for the the element named 'loc' in any namespace
+    urls = xml.xpath("//*[local-name()='loc']/text()")
+
+    # THEN non-OpenStax derived copies have been removed
+    # while OpenStax books remain
+    found_url = [url for url in urls if uuid in url]
+    assert xnor(should_be_present, found_url)
```
pumazi: @m1yag1 do you feel this is better or worse? 😬

pumazi: https://en.wikipedia.org/wiki/XNOR_gate

m1yag1: hmm.. i might have to run it manually but based on the truth table.
`found_url = True`
`should_be_present = False`
`xnor(should_be_present, found_url)` == True. 

Wouldn't we want `should_be_present` False to fail if there was a found_url?

Also the opposite, if should_be_present is True and found urls is False we would get True.

I could also not be understanding exactly what's going on here. That is entirely possible as it's after 2pm when my brain goes retarded.

m1yag1: I'll run it manually. i do like the approach tho. it's more readable and uses boolean logic 👍

m1yag1: Actually i just noticed my mistake... haha. i completely didn't see the not in the xor function derp

m1yag1: can you add the wiki link as a comment somewhere?

m1yag1: can say something like #see truth table: https://en.wikipedia.org/wiki/XNOR_gate

m1yag1:
BTW, On the research team the readings we used to test learning and retention involved a reading of digital logic. It was pretty cool and a lot of the participants sent gave feedback about how they had always heard of the topic but never learned it. They were glad to have the opportunity to actually learn it and get tested on it. ref:  https://github.com/openstax/research-p3-experiment-4/blob/master/digital_logic/templates/textbook/digital_logic.md

m1yag1:
Note: Approved but am waiting on response if we should add xfail. Please ping me on the result.

PR #632, opened 2019-12-05
changed donate url to the current new one

omehes:
test_donate.py failed because the donate url changed. new address is https://riceconnect.rice.edu/donation/support-openstax


omehes:
when I re-run the failed test 'test_share_links_displayed' it passed on cnx.org but it fails on staging.cnx.org as there is a slight diff in linkedin url. not related to my change in test_donate.

PR #633, opened 2019-12-06
added test case to verify robots on production (and archive.cnx.org)

omehes:
verifying current /robots.txt against a /robots.txt from cache from nov 27: https://webcache.googleusercontent.com/search?q=cache:7VQ6FHq3dHUJ:cnx.org/robots.txt+&cd=1&hl=en&ct=clnk&gl=us


omehes:
The first 2 fails are known and will be updated.
The third fails (91) is this test case being implemented. It fails because circleci runs  staging.cnx.org and test_robots.prod.py is for cnx.org. When webview_base_url is staging.cnx.org, it will fail. Can it be skipped?

m1yag1:
Otto there is a fixture that provides  the instance from the webview_base_url. For example for https://qa.cnx.org it will return qa. If it doesn't detect a subdomain it will return 'prod'.

https://github.com/openstax/cnx-automation/blob/master/fixtures/webview.py#L301-L307

You can use this fixture to a skip a test using pytest.skip. 
http://doc.pytest.org/en/latest/skipping.html#skipping-test-functions

You could also write 2 separate tests for each instance.

omehes:
@m1yag1 I worked on the robots tests, created 2 test, one for staging and skipping anything else, and for prod, skipping anything else. I tried it out and they mark testrail either pass or blocked.

tests/webview/ui/test_robots_staging.py
```diff
@@ -23,13 +25,18 @@ def test_robots(webview_base_url, archive_base_url, selenium):
...
         assert "\nUser-agent: ScoutJet\nCrawl-delay: 10\nDisallow: /\n" in webview_robots_text
         assert "\nUser-agent: Baiduspider\nCrawl-delay: 10\nDisallow: /\n" in webview_robots_text
         assert "\nUser-agent: BecomeBot\nCrawl-delay: 20\nDisallow: /\n" in webview_robots_text
         assert "\nUser-agent: Slurp\nCrawl-delay: 10\nDisallow: /\n" in webview_robots_text
 
-    assert "\nUser-agent: *\nDisallow: /\n" in archive_robots_text
+        assert "\nUser-agent: *\nDisallow: /\n" in archive_robots_text
+
+    else:
+        print("\n\nTesting staging only, skipping other environments\n")
```
m1yag1: Remove print statement

tests/webview/ui/test_robots_production.py
```diff
@@ -0,0 +1,38 @@
...
+    if "//cnx.org" in webview_base_url:
+
+        with open(robots_production_list, "r") as file:
+            readf = "\n" + file.read()
+
+            assert readf == webview_robots_text
+            assert "\nUser-agent: *\nAllow: /\n" in archive_robots_text
+
+    else:
+        print("\n\nTesting production only, skipping other environments\n")
```
m1yag1: Remove print. Proper naming of variables and code should suffice.

omehes:
@m1yag1 changes done

PR #634, opened 2019-12-16
verifying vendor books and pages that they don't redirect

omehes:
verifying that vendor book titles don't redirect
verifying that vendor book/pages don't redirect
and all associated files for the above test files


omehes:
it is the donate page which fails as I forgot to update the branch

omehes:
I am closing this PR and will create a new one.

PR #636, opened 2019-12-16
verifies that all rex redirects are permanent with code 301

omehes:
verifies that all rex books and pages redirects permanently (301 code)


tests/webview/integration/test_redirects_301.py
```diff
@@ -2,136 +2,77 @@
...
     # WHEN we go to a page based on the webview_base_url and uri
-    cnx_page_slug = chemistry_2e_uri.split("/")[-1]
     cnx_url = f"{webview_base_url}{chemistry_2e_uri}"
-    response = get_url(cnx_url)
+    response = requests.get(cnx_url, allow_redirects=False)
 
     # THEN we are redirected to rex
-    assert response.url.startswith(rex_base_url)
-    assert response.url.endswith(cnx_page_slug)
+    print()
```
m1yag1: print statements are suppressed by the test framework unless you use `-s` or the test fails. Generally, printing should be avoided in tests. These print statements are also confirmed by your assert statements. Please get rid of them. Additionally, the test framework is set to `showlocals` which will print out any local variables. if you need insight to why a test failed you can set a value to a variable.

PR #638, opened 2020-01-06
readding two test files as they were somehow deleted 

omehes:
Somehow I deleted these 2 files on the redirects_as_301 branch which I just merged to master. Never noticed this and now these 2 files are missing on master. I am readding them here. No changes just minor changes in the comments of both files to trigger change.


m1yag1:
PR not necessary anymore.

PR #647, opened 2020-02-13
New 9 books to redirect to openstax.org (with eship)

omehes:
Economics 2e
Microeconomics 2e
Macroeconomics 2e
Entrepreneurship
Sociology 2e
Intro Business
Business Ethics
Principles of Management
Organizational Behavior


omehes:
@m1yag1 These fails are related to new redirecting books and we will have more and more of them. Some test we will have to switch off soon on staging.cnx.org and cnx.org

PR #652, opened 2020-03-13
add test that archive redirects respect host header

TomWoodward:
Test should run in cnx-automation by using:

```shell
pytest -m archive --webview-base-url https://qa.cnx.org --legacy-base-url https://legacy-qa.cnx.org --archive-base-url https://archive-qa.cnx.org
```

### DIFF
--- a/tests/webview/integration/test_archive_redirects.py
+++ b/tests/webview/integration/test_archive_redirects.py
@@ -1,0 +1,15 @@
+import requests
+
+from tests import markers
+
+@markers.nondestructive
+@markers.parametrize("uuid", [("fea3130c-6e57-41b2-a00f-0267ffae273c")])
+def test_archive_redirect(archive_base_url, uuid):
+    # GIVEN an "archive" URL
+    url = f"{archive_base_url}/contents/{uuid}"
+
+    # WHEN a web request has a host
+    response = requests.get(url=url, headers={"Host": "openstax.org"}, allow_redirects=False)
+
+    # THEN the host is used in the redirect 
+    assert "openstax.org" in response.headers["Location"]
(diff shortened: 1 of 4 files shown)


tests/webview/integration/test_archive_redirects.py
```diff
@@ -0,0 +1,15 @@
+import requests
+
+from tests import markers
+
+@markers.nondestructive
+@markers.parametrize("uuid", [("fea3130c-6e57-41b2-a00f-0267ffae273c")])
+def test_archive_redirect(archive_base_url, uuid):
```
m1yag1: It's a bit unclear to me what this test is actually testing based on the name.

Would a better test name be `test_archive_redirect_contains_openstax_host_header`? 

We're completely fine with long test names. The idea is that they clearly convey the intent of the test.

Perhaps, the test itself could be made more generic and the header be included as a parameter? In this case, the test would become `test_archive_redirect_headers` or something like that.

tests/webview/integration/test_archive_redirects.py
```diff
@@ -0,0 +1,15 @@
...
+    # GIVEN an "archive" URL
+    url = f"{archive_base_url}/contents/{uuid}"
+
+    # WHEN a web request has a host
+    response = requests.get(url=url, headers={"Host": "openstax.org"}, allow_redirects=False)
```
m1yag1: I think parametrizing the header would be good here.

TomWoodward:
@m1yag1 i based my test off of https://github.com/openstax/cnx-automation/blob/master/tests/webview/integration/test_webview_integration.py, but if you parameterize the headers you might be able to combine the tests. the only difference is that my test relies on the version not being set in the request

PR #655, opened 2020-03-19
auto-test: new redirecting book - prealgebra 2e

omehes:
This is to test the latest book, Prealgebra 2e, added to the list of redirecting books


omehes:
Fail not related.

PR #656, opened 2020-03-25
Updated README.md with steps to create redirects data

omehes:
added section for creating redirects data - uuid.txt files containing url variants


omehes:
Fail is not related

README.md
```diff
@@ -238,3 +238,18 @@ pattern.
 [pytest]: https://docs.pytest.org/en/latest/
 [mozilla]: https://github.com/mozilla/addons-server
 [selenium-training]: https://qualitas-server.herokuapp.com/wiki/Introduction_to_Pytest_and_Selenium
+
+### HOW TO CREATE TEST DATA FOR REDIRECTS
+
+1. go to ce-scripts repo
+2. cd python
```
rnathuji: Perhaps we should add a note to create a `venv` and run `pip install -r requirements.txt` here prior to running `gen_book_uris.py`?

README.md
```diff
@@ -238,3 +238,18 @@ pattern.
...
+3. run gen_book_uris.py <archive_host> <cnx_uuid>
+e.g. $ ./gen_book_uris.py archive-staging.cnx.org e42bd376-624b-4c0f-972f-e0c57998e765
+4. this will create a cnx_uuid.txt file with all url variations as per rex_redirects.py
```
rnathuji: Is this correct? I see the file generated as `output/<uuid>.txt`. So, for example, after running the command above I get a `output/e42bd376-624b-4c0f-972f-e0c57998e765.txt` file.

README.md
```diff
@@ -238,3 +238,18 @@ pattern.
...
 [selenium-training]: https://qualitas-server.herokuapp.com/wiki/Introduction_to_Pytest_and_Selenium
+
+### HOW TO CREATE TEST DATA FOR REDIRECTS
+
+1. go to ce-scripts repo
+2. cd python
+3. run gen_book_uris.py <archive_host> <cnx_uuid>
+e.g. $ ./gen_book_uris.py archive-staging.cnx.org e42bd376-624b-4c0f-972f-e0c57998e765
+4. this will create a cnx_uuid.txt file with all url variations as per rex_redirects.py 
+5. this file can be found here: /usr/local/lib/python3.7/site-packages/rex_redirects.py
```
rnathuji: Is this just an informative point as far as what the underlying code for `rex_redirects.py` is, or is it actionable? If the former, perhaps we can link to `https://github.com/openstax/cnx-rex-redirects/blob/master/rex_redirects.py` in this doc instead?

README.md
omehes: OK

README.md
omehes: Yes. It is output. I will add it.

README.md
omehes: Informative point to know where that code is.

omehes:
@rnathuji I made the changes as per your review

m1yag1:
Why not add these instructions to the main README.md file?

README.md
```diff
@@ -238,3 +238,41 @@ pattern.
 [pytest]: https://docs.pytest.org/en/latest/
 [mozilla]: https://github.com/mozilla/addons-server
 [selenium-training]: https://qualitas-server.herokuapp.com/wiki/Introduction_to_Pytest_and_Selenium
+
+### HOW TO CREATE TEST DATA FOR REDIRECTS
```
rnathuji: Sorry,  I should have caught this the first time, but I just now looked at it in rich diff form: In terms of structure should this be ## vs ###? In other words, is this section a child of the "Framework Design" section or a peer?

Also, just a minor comment, but perhaps it would be nice to line up the styling with the other sections so they match a bit better (?) (this is a more subjective point, I know)

README.md
```diff
@@ -238,3 +238,41 @@ pattern.
...
+
+```gen_book_uris.py <archive_host> <cnx_uuid>```
+
+e.g. $ ./gen_book_uris.py archive-staging.cnx.org e42bd376-624b-4c0f-972f-e0c57998e765
+
+this will create url variations as per rex_redirects.py and stores it in output/<cnx_uuid.txt> file 
+
+e.g. output/e42bd376-624b-4c0f-972f-e0c57998e765.txt
+
+rex_redirects.py file can be found here: /usr/local/lib/python3.7/site-packages/rex_redirects.py
```
rnathuji: I think using `/usr/local/lib/python3.7/site-packages/rex_redirects.py` explicitly like this is problematic since the exact path would be environment specific. Why not just embed a link to https://github.com/openstax/cnx-rex-redirects/blob/master/rex_redirects.py so readers can follow the reference that way?

omehes: @rnathuji updated the readme file with the link.

README.md
omehes: updated the title height with ##

omehes:
@rnathuji I updated the layout a bit and also updated the branch with the latest changes in master branch.

PR #650, opened 2020-02-26
e2e test of cops jobs

omehes:

PR #657, opened 2020-03-25
auto-test: updated cops e2e test file

omehes:
updated the cops e2e test file:
- rewrote the verification code
- added content check for the created pdf doc

### DIFF
--- a/fixtures/cops.py
+++ b/fixtures/cops.py
@@ -1,0 +1,82 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import os
+import pytest
+import requests
+
+import tldextract
+from rex_redirects import get_rex_release_json_url
+
+from tests.utils import gen_from_file, skip_if_destructive_and_sensitive
+
+DATA_DIR = os.path.join(os.path.realpath(os.path.dirname(__file__)), "data", "webview")
+
+__all__ = ["cops_base_url"]
+
+
+@pytest.fixture
+def cops_base_url(request):
+    """Return a base URL for cops"""
+    config = request.config
+    base_url = config.getoption("cops_base_url") or config.getini("cops_base_url")
+    if base_url is not None:
+        skip_if_destructive_and_sensitive(request, base_url)
+        return base_url
+
+
+@pytest.fixture
+def s3_base_url(request):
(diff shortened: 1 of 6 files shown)


omehes:
created this new PR and closed the original one: https://github.com/openstax/cnx-automation/pull/650/commits

omehes:
New changes to cops_e2e and cops_ui tests:
1fefc23 (HEAD -> cops_e2e_pdf, origin/cops_e2e_pdf) removed cops and s3 books fixtures from fixtures/webview.py and added new fixtures/cops.py file to include cops and s3 fixtures
36de97d updated conftest.py and setup.cfg with cops_base_url
acf5c77 new test to verify cops ui
afbf90d updated conftest.py with cops_base_url
50b3100 updated webview.py with cops_base_url
52248ed updated webview.py with cops_e2e fixtures

fixtures/cops.py
```diff
@@ -0,0 +1,82 @@
...
+import os
+import pytest
+import requests
+
+import tldextract
+from rex_redirects import get_rex_release_json_url
+
+from tests.utils import gen_from_file, skip_if_destructive_and_sensitive
+
+DATA_DIR = os.path.join(os.path.realpath(os.path.dirname(__file__)), "data", "webview")
```
rnathuji: Since you're splitting out the fixtures anyways (which makes sense), should the data files `s3_books_uuids.txt` and `s3_books_titles.txt` perhaps live in `data/cops` instead of `data/webview`?

fixtures/cops.py
```diff
@@ -0,0 +1,82 @@
...
+__all__ = ["cops_base_url"]
+
+
+@pytest.fixture
+def cops_base_url(request):
+    """Return a base URL for cops"""
+    config = request.config
+    base_url = config.getoption("cops_base_url") or config.getini("cops_base_url")
+    if base_url is not None:
+        skip_if_destructive_and_sensitive(request, base_url)
```
rnathuji: I may be wrong or misunderstanding how this works, but will this skipping ever get invoked? The `sensitive_url=^(?:https?://)?(?:archive\.|legacy\.)?cnx\.org` regex in `setup.cfg` will never match `https://cops.openstax.org/` for example. Is this something we want to address?

omehes: That's a good point. I just copied this from qwebview and used for cops.

fixtures/cops.py
omehes: Good point. Didn't think of it.

PR #658, opened 2020-03-26
Move archive tests over to new application folder

m1yag1:
It makes more sense to have things that are testing archive directly like redirects to be in their respective folder. 

This PR needs to be merged in first!!!: #652


m1yag1:
This PR needs to be merged in first!!!: https://github.com/openstax/cnx-automation/pull/652

PR #665, opened 2020-04-02
fixing hanging collection id in test_content

omehes:
previous collection:page short ids failed and remained hanging. changed it to 
pytest.param("TqqPA4io:_LheuYk4") in
test_content:test_chapter_review_version_matches_book_version


omehes:
@m1yag1 Fail not related and passes when run separately.

PR #667, opened 2020-04-17
will not update chrome driver

c0ldcalamari:
testing PR to see if I made the correct update for tests to pass, to get an idea what update work may look like.


m1yag1:
I don't think this is necessary since we have another pyup PR that has these updates.

PR #669, opened 2020-04-22
Fix stale element errors on the homepage

m1yag1:


m1yag1:
PR not needed as we'll be removing webview UI test code.

PR #678, opened 2020-05-15
changed book index in test_content due to new redirecting books

omehes:
had to change book list index from 0 to 1 as more and more books redirects.
Also, flake8 didn't like that there was no placeholder in f"... string, so I removed f

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,8 +5,11 @@ repos:
     rev: stable
     hooks:
       - id: black
-        language_version: python3.7
+        language_version: python3.8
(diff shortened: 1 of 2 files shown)


therealmarv:
looks okay. Only question: Why change the index from 0 to 1 ? Feels like a logic change... just want to know the context.

omehes:
@therealmarv first book (index 0) is Prealgebra 2e and now it redirects. So, the tests in test_content fail. Next book not redirecting is Prealgebra (index 1) on staging. Content is different on qa but staging and prod have similar content.

PR #680, opened 2020-06-17
Bn/cnx#910/neb archive regression

c0ldcalamari:
Need clarification/understanding on `@markers.test_case("C282983")`

tests/neb/ui/test_get.py:131

```
@markers.neb
@markers.test_case("C282983")
@markers.nondestructive
```


requirements.txt
```diff
@@ -37,11 +37,11 @@ idna==2.7                 # via -r requirements.in, requests, tldextract, yarl
 importlib-metadata==0.23  # via -r requirements.in, pluggy, pre-commit, pytest, virtualenv
 jinja2==2.11.2            # via -r requirements.in, cnx-epub
 jwcrypto==0.6.0           # via -r requirements.in, github3.py
-lxml==4.5.0               # via -r requirements.in, cnx-epub, cnx-litezip, cnxml, rhaptos.cnxmlutils
+lxml==4.4.3               # via -r requirements.in, cnx-epub, cnx-litezip, cnxml, rhaptos.cnxmlutils
 markupsafe==1.1.1         # via -r requirements.in, jinja2
 more-itertools==8.2.0     # via -r requirements.in, pytest
 multidict==4.7.5          # via aiohttp, yarl
-nebuchadnezzar==9.6.0     # via -r requirements.in
+git+https://github.com/openstax/nebuchadnezzar.git
```
rnathuji, line 44: I'm not sure if these changes are intentional for this PR or an accidental inclusion, but in general I believe the `requirements.txt` file is [auto-generated in this repo](https://github.com/openstax/cnx-automation#doing-updates-to-dependencies). If there are proposed dependency version changes that need to be made (?), they should be reflected in a patch to `requirements.in`.

cli/neb.py
```diff
@@ -67,9 +67,12 @@ def get(cls, *, help=False, verbose=False, env=None, col_id=None, col_version=No
                 # because neb refuses to use a dir that already exists
                 neb_dir = join(temp_dir, "neb")
 
-                options = ["--output-dir", neb_dir]
+                options = ["--output-dir", neb_dir, "--insecure"]
```
rnathuji, line 70: While I think it's useful to be able to run these tests against a local VM, I'm wondering if we want to include `--insecure` all of the time?

One alternative approach to consider could be making it configurable similar to [`neb_env`](https://github.com/openstax/cnx-automation/blob/f7ece379b32335ae90d05dd6d937713d83ffb268/setup.cfg#L5) (e.g. something like `neb_insecure`) and pipe through.

tests/neb/ui/test_get.py
```diff
@@ -126,3 +126,20 @@ def test_get_col_minor_version(neb_env, col_id, col_version, snapshot):
     ) as zip_dir:
         # THEN the complete zip is downloaded and matches the snapshot
         snapshot.assert_file_or_dir_match(zip_dir, snapshot_name)
+
+@markers.neb
+@markers.nondestructive
+@markers.parametrize("col_id,col_version", [("col11562", "1.19.6")])
+def test_get_col_with_resources(neb_env, col_id, col_version, snapshot):
```
rnathuji, line 133: I'm curious about this test case. I understand that it differs from the preceding test cases in that it also downloads the resource files. However, does this exercise a new code path on archive (I'm assuming that's the intent -- if the goal is to exercise a new code path on `neb`, then ignore this comment)? 

Specifically, I believe `neb` will use [`/resources/{hash}`](https://github.com/openstax/cnx-archive/blob/027dd59d5ae0dc4045945ea416269487929d550a/cnxarchive/__init__.py#L66) for both [content files](https://github.com/openstax/nebuchadnezzar/blob/6aa1d2d3dbc260b82c38f402623c784506a9f85b/nebu/cli/get.py#L306) and  [resource files](https://github.com/openstax/nebuchadnezzar/blob/6aa1d2d3dbc260b82c38f402623c784506a9f85b/nebu/cli/get.py#L326). In archive, I believe both types of invocations lead to the [same code being exercised](https://github.com/openstax/cnx-archive/blob/027dd59d5ae0dc4045945ea416269487929d550a/cnxarchive/views/resource.py#L32). Am I missing something?

tests/neb/ui/test_get.py
```diff
@@ -126,3 +126,20 @@ def test_get_col_minor_version(neb_env, col_id, col_version, snapshot):
...
+    # GIVEN neb, an environment name, a collection id, a collection version, and the snapshot tool
+    snapshot_name = get_neb_snapshot_name(col_id, col_version, "resources")
+
+    # WHEN we run `neb get --verbose env col_id col_version`
+    with Neb.get(
+        verbose=True, resources=True, env=neb_env, col_id=col_id, col_version=col_version, input="y"
+    ) as zip_dir:
+        # THEN the complete zip is downloaded and matches the snapshot
+        snapshot.assert_file_or_dir_match(zip_dir, snapshot_name)
```
rnathuji, line 142: Does this need a [snapshot file](https://github.com/openstax/cnx-automation/tree/master/snapshots/tar_gz/neb/col11562) to be added for comparison?

I noticed `test_get_col_minor_version` doesn't have a corresponding `1.19.1.tar.gz` file (and instead the file gets written out through [this code path](https://github.com/openstax/cnx-automation/blob/f7ece379b32335ae90d05dd6d937713d83ffb268/fixtures/snapshot.py#L74)). I'm not sure if that's intentional for that test, but I suspect you may want to compare against a known snapshot for this one (?)

tests/neb/ui/test_get.py
```diff
@@ -126,3 +126,20 @@ def test_get_col_minor_version(neb_env, col_id, col_version, snapshot):
     ) as zip_dir:
         # THEN the complete zip is downloaded and matches the snapshot
         snapshot.assert_file_or_dir_match(zip_dir, snapshot_name)
+
+@markers.neb
+@markers.nondestructive
+@markers.parametrize("col_id,col_version", [("col11562", "1.19.6")])
+def test_get_col_with_resources(neb_env, col_id, col_version, snapshot):
+    # GIVEN neb, an environment name, a collection id, a collection version, and the snapshot tool
+    snapshot_name = get_neb_snapshot_name(col_id, col_version, "resources")
```
rnathuji, line 135: Is the the use of this additional `tag` component necessary since you're using a unique version from other tests? (e.g. the `1.19.6.tar.gz` would already be unique from `1.19.1.tar.gz`)

tests/neb/ui/test_get.py
```diff
@@ -126,3 +126,20 @@ def test_get_col_minor_version(neb_env, col_id, col_version, snapshot):
...
+
+    # WHEN we run `neb get --verbose env col_id col_version`
+    with Neb.get(
+        verbose=True, resources=True, env=neb_env, col_id=col_id, col_version=col_version, input="y"
+    ) as zip_dir:
+        # THEN the complete zip is downloaded and matches the snapshot
+        snapshot.assert_file_or_dir_match(zip_dir, snapshot_name)
+
+
+    # first pass will return a false negative so  should test be ran twice?
```
rnathuji, line 145: Minor: Not sure if this comment is still valid or an accidental remnant

cli/neb.py
tomjw64: If this is a true regression test, wouldn't it be hard to run it against anything _other_ than a VM? Running it against something live and with the potential to change in-between runs seems futile for the purposes of testing for regressions. Maybe my understanding is different since I haven't visited this work in a while. @c0ldcalamari - how are these tests expected to be run?

rnathuji: My understanding (which may be incorrect) is that these existing neb tests are run against QA and (perhaps) other environments. I'm not sure about the periodicity or if that's automated somewhere, but running them against a VM with self-signed certs is something new. Not checking the certs isn't going to necessarily hurt those tests, but similar to how we changed `neb` to ignore cert checking only when we know we need that behavior seems like the right way to go versus always ignoring them, no? 🤷‍♂️

fixtures/neb.py
```diff
@@ -14,3 +14,11 @@ def neb_env(request):
...
     if neb_env is not None:
         return neb_env
+
+@pytest.fixture(scope="session")
+def insecure(request):
+    """Returns the insecure flag setting"""
+    config = request.config
+    insecure = config.getoption("insecure")
+    if insecure is not None:
+        return insecure
```
tomjw64, line 24:
```suggestion
    return insecure
```
If insecure is `None`, the function will not explicitly return, which implicitly returns `None`. This is the same as just returning the value whether it is `None` or not.

Alternatively,
```suggestion
    if insecure is not None:
        return insecure
    return False
```
might be neater depending on your viewpoint.

PR #688, opened 2020-07-21
Fix assert to force stdout to string

rnathuji:
This addresses an issue with the modified test failing in python 3.8.


omehes:
@rnathuji verified. with the fix no errors when running pytest neb_publish. looks good

PR #691, opened 2020-08-11
adding boto3 to requirements file and updated chrome-driver version

omehes:


omehes:
@c0ldcalamari added the boto3 as per recommendation from Mike. Also updated chromedriver.

PR #693, opened 2020-09-09
added fixtures for code_tag and queue_state_bucket_name

omehes:
s3 bucket book verification: added fixtures for code_tag and queue_state_bucket_name, so when they change, we can supply them as command lines via pytest

### DIFF
--- a/fixtures/webview.py
+++ b/fixtures/webview.py
@@ -623,3 +623,25 @@ def aws_secret_access_key_value(request):
     awssecretvalue = config.getoption("aws_secret_access_key_value")
     if awssecretvalue is not None:
         return awssecretvalue
+
+
+@pytest.fixture
+def code_tag(request):
+    """Return a deployment code tag"""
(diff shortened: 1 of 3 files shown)


fixtures/webview.py
```diff
@@ -623,3 +623,25 @@ def aws_secret_access_key_value(request):
...
         return awssecretvalue
+
+
+@pytest.fixture
+def code_tag(request):
+    """Return a deployment code tag"""
+    config = request.config
+    code_tag = config.getoption("code_tag") or config.getini("code_tag")
+    if code_tag is not None:
+        skip_if_destructive_and_sensitive(request, code_tag)
```
rnathuji, line 634: Will ` skip_if_destructive_and_sensitive` ever do anything in these two fixtures given it looks for specific URL structures (assuming I understand it correctly)?

rnathuji (approved):

PR #699, opened 2020-10-07
adding code_tag instead of fixed master, modifying exception handling

omehes:
as per discussion with Ripal, I modified the s3 bucket book tests by adding code_tag instead of fixed master, modifying exception handling in the fixture and some comments


fixtures/bucket_books_tree.py
```diff
@@ -39,13 +41,27 @@ def bucket_books_tree(s3_base_url, s3_queue_state_bucket_books):
         f"{s3_base_url}{s3_archive_folder}{i}@{j}.json" for i, j in zip(uuid_list, version_list)
     ]
 
-    book_tree = []
+    # list of approved book urls without http errors
+    approved_books_full_url_modified = []
+
+    for ublu in approved_books_full_url:
```
rnathuji, line 47: This is really a comment on line 25 above (GH won't let me comment on that line, though 🤷‍♂️ ): Is that effectively de-duplicating by the `name` field? If so, won't that prevent checking multiple versions for a given book?

How many successful books this test currently finds for our recent staging pipeline? I believe it should be 85 successes and 28 failures (out of 113 books and 141 queue state entries).

fixtures/bucket_books_tree.py
```diff
@@ -39,13 +41,27 @@ def bucket_books_tree(s3_base_url, s3_queue_state_bucket_books):
...
+        else:
+            # Return code 200
+            approved_books_full_url_modified.append(ublu)
+
+    book_tree = []
 
-        s3_books = urllib.request.urlopen(approved_books_full_url[books]).read()
+    for books in range(len(approved_books_full_url_modified)):
 
+        s3_books = urllib.request.urlopen(approved_books_full_url_modified[books]).read()
```
rnathuji, line 62: Minor: I believe you can insert the results into `book_tree` in the `try` block while iterating on `approved_books_full_url` instead of using `approved_books_full_url_modified` as an intermediate list you iterate on in a second pass.

fixtures/bucket_books_tree.py
omehes: @rnathuji I see. Yes, we put this code there with Brenda to weed out duplicates but now the situation changed somewhat, so it is not necessary to find duplicate books (by title). I removed that bit of code, no more checking for duplicates, all books fall trough the filter :)

fixtures/bucket_books_tree.py
omehes: @rnathuji I will have a look at this later once I am done with tests. For now I pushed changes regarding your first comment.

rnathuji (approved):

PR #700, opened 2020-10-13
New redirecting books and updated ones

omehes:
Adding 2 new redirecting books
Updated books with Ripal's fixes
Separated out these latest books into a second set of redirect tests as the amount of test data is too huge and time to test too long


rnathuji:
@omehes - I think you may have unintentionally added a couple of unrelated commits to the PR.

rnathuji (approved):

PR #702, opened 2020-10-19
removing duplicate title check and separating exception check

omehes:
I removed duplicate title check in another fixture and separated exceptions


omehes:
I did some more changes to have less messaging when tests are running and only see assert errors and http errors (when concourse jobs fail).

rnathuji (approved):

PR #703, opened 2020-10-20
changed test code to use xhtml files and more

omehes:
after our discussion yesterday, I thought why not to add verification of book uuids between queue state file and books in the bucket.


omehes:
sorry @rnathuji I added verification of uuids. This should be it :) unless you think that checking book titles is an utter waste of time...

tests/s3_bucket/test_s3_pages_content.py
```diff
@@ -75,21 +77,31 @@ def test_page_content(bucket_books_tree, s3_books_title_version_dict, code_tag,
...
 
         book_title.clear()
         book_title.append(bucket_books_tree[elem]["title"])
 
         title_results = list(s3_books_title_version_dict.keys())
 
+        try:
+
+            # compares book uuids in the queue-state list against actual books in aws s3 bucket
+            assert [j for j in book_id if j in s3_books_uuids]
```
rnathuji, line 91: Just to confirm my understanding: The goal of this assertion is to fail if a book UUID is in the bucket but not in the queue state list, correct? If so, I think the problem is `bucket_books_tree` itself is generated based upon `s3_queue_state_bucket_books`, so is it even possible for this to happen in terms of the code? (e.g. I think to really check for this you'd have to do an `s3 ls` equivalent on the bucket)

tests/s3_bucket/test_s3_pages_content.py
```diff
@@ -75,21 +77,31 @@ def test_page_content(bucket_books_tree, s3_books_title_version_dict, code_tag,
...
+
         try:
 
             # compares book titles in the queue-state list against actual books in aws s3 bucket
             assert [i for i in book_title if i in title_results]
 
         except AssertionError as as_error:
 
-            print(f"{as_error}: book title {book_title} not found, next item")
+            print(f"{as_error}: book title {book_title} : {book_id} not found, next item")
```
rnathuji, line 104: To your question, I do think this is probably a meaningless check. If it fails, it means we won't check pages for a book that exists because it's in the ABL, and the fact that the titles don't match really doesn't matter. Perhaps remove this assertion check? 🤷‍♂️

rnathuji:
@omehes - I think there may also be a problem with some of the code [here](https://github.com/openstax/cnx-automation/blob/8347bae5cae1a6e06b634135d8dad75e0f1dbb17/tests/s3_bucket/test_s3_pages_content.py#L117-L123) because the generated URLs don't include book version I believe. So, for example, it will try to pull https://release-1342.sandbox.openstax.org/apps/archive/20201005.151934/contents/27f59064-990e-48f1-b604-5188b9086c29:e4f332ca-e844-5f3d-b7a4-f1c080506983.json (which doesn't work) instead of https://release-1342.sandbox.openstax.org/apps/archive/20201005.151934/contents/27f59064-990e-48f1-b604-5188b9086c29@12.1:e4f332ca-e844-5f3d-b7a4-f1c080506983.json.

omehes:
@rnathuji after our discussions I changed the code to utilise xhtml files to get all the page urls. Removed some try/except code and more.

tests/s3_bucket/test_s3_bucket_books.py
```diff
@@ -0,0 +1,136 @@
...
+        # verifies every 10th page url in each book
+        for link in links[::10]:
+
+            links_replaced = link.replace("./", f"{s3_base_url}{s3_archive_folder}").replace(
+                ".xhtml", ".json"
+            )
+
+            res = requests.get(links_replaced)
+
+            s3_pages = urllib.request.urlopen(links_replaced).read()
```
rnathuji, line 128: Minor: I think you're requesting the same link twice here, no? Could you remove this line and make the next one `s3_pages_jdata = res.json()`?

rnathuji (approved):

tests/s3_bucket/test_s3_bucket_books.py
omehes: @rnathuji I will have a look at it together with other changes I started doing. just to clean up the fixtures and code. I merge this and I will come back to this after redirects tests.

omehes:
@rnathuji so, here are the additional changes I madeas I realised that some fixtures are not needed anymore and I cleaned up the test code. I have not looked into your comment yet, I will come back to it once I am done with redirects tests.

omehes:
@rnathuji another commit to the same branch, I fixed the request code duplication

rnathuji:
@omehes - This PR looks good to me to merge.

PR #717, opened 2020-12-15
qa testing tool to verify version changes at deployments (between environs qa and staging against prod)

omehes:
this is a tool for qa tester to check on version changes in cnx, when we deploy/release new build to qa and staging


rnathuji:
I have a general question on the tests added with this PR: How (if at all) do they relate to the [Qualitas version dashboard](https://qualitas-server.herokuapp.com/dashboards/cnx-repos)?

m1yag1:
These tests also don't assert anything.

omehes:
They relate somewhat, I guess in that the test checks similar/same sources for versions. I created this test as it simplifies the version change checks for me. I would run the test at every deploy and as an output, I get only changed values, like this:
```tests/versions_updates/test_versions_updates.py::test_versions_updates 

('values_changed', 2)
("root['cnx-recipes']", {'new_value': '1.67.0', 'old_value': '1.66.0'})
("root['importlib-metadata']", {'new_value': '0.19', 'old_value': '0.21'})
("root['cnx-common']", {'new_value': '1.2.1', 'old_value': '1.2.2'})
("root['cnx-transforms']", {'new_value': '1.0.1', 'old_value': '1.1.0'})
PASSED
tests/versions_updates/test_versions_updates.py::test_webview_versions_updates 

("root['date']", {'new_value': '2020-12-08 11:42:41 CST', 'old_value': '2020-12-07 15:00:01 CST'})
("root['cnx-deploy']", {'new_value': '20201208.163442', 'old_value': '20201203.163400'})
PASSED
```
It makes checking simpler and faster for me and I can message these results in the channel, to catch probable version issues, just like we had with cnx-recipes. I used to do this before but manually but after some updates to deploy process, I stopped...

m1yag1:
Please see my previous comment for the diff tool. I can make some updates to the diff tool so that things are done via query params and that way you can share the url.

omehes:
I initially had assert in it and could be added, but then I added print(). Ok, this is a tool for me rather than for automation. I just wanted for someone to have a look at it and see if there is any flaw in it. But I guees it is waste of time. I got a bit carried away :) It makes verification of version changes easier and much fasteer for me.

m1yag1:
I wouldn't say it's a waste of time if it helps you. There is also not anything wrong with getting feedback. It seems like a valuable tool but maybe not the right fit in this framework. Like I mentioned earlier you may be able to add some warnings b/c it's hard to make a determination if a version change is actually a failure. That's why the version dashboard displays things in red b/c a human has to verify if something actually needs an update.

omehes:
No, no, the feedback is great. That is why I created this PR :) I will see what assertion can do here.

m1yag1:
@omehes btw, there are objects in the repository that parse the history.txt

Example: 
https://github.com/openstax/cnx-automation/blob/master/parsers/cnx_deploy/release.py#L10-L42

We print a warning if the version of neb is out of date for the tests we are running:
https://github.com/openstax/cnx-automation/blob/master/tests/neb/ui/test_usage_and_version.py#L56-L82

You should probably try to adapt your code to adopt these objects as that's what they are designed for. The complexity is abstracted away and your test can read like the one above.

brittweinstein:
@omehes Please make this a card and attach.

omehes:
@m1yag1 hey. I am reading through your comments, checking the included links but I think I need more help. Not sure how/where to start. Do you have any more pointers for me?

PR #720, opened 2021-01-28
improvements and amendments to neb publish tests

omehes:
- trying to minimise intermittent failures
Note, that tests/neb/ui/test_publish.py::test_successful_publish[col11562-latest-Publication message from cnx-automation] is still seems to fail. Will create another PR for that.


omehes:
For some reason, tests/neb/ui/test_get.py::test_get_col_latest[col11562-1.19] fails on some environs, like staging. This is independent of my changes byt will investigate.

omehes:
I did another commit to change col ID in get publish test too (from col11562 to col11496) as for some reason, col11562 is giving fails when used with "latest".

rnathuji (approved):

PR #725, opened 2021-03-17
test modules of git content repo

omehes:


fixtures/github_authorization.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+
+
+@pytest.fixture
+def github_authorization(github_tok):
+    """Returns the authorization token for github
+    """
+    headers = {"Authorization": "token " + github_tok, "Accept": "application/vnd.github.v3.raw"}
```
rnathuji, line 8: Do you want to separate the `Accept` header from the `Authorization` or perhaps rename the fixture since the former is separate from authorization??

tests/s3_bucket/test_github_content_repos.py
```diff
@@ -0,0 +1,52 @@
...
+            modules_url = (
+                f"https://api.github.com/repos/openstax/{repo}/contents/{rel_path}/index.cnxml"
+            )
+
+            modules_req = urllib.request.Request(modules_url, headers=github_authorization)
+            module_resp = urllib.request.urlopen(modules_req)
+
+            if module_resp.status != 200:
+                # Assert here if expected index.cnxml is not found
+                print(f"FAILED to find index.cnxml in {rel_path}")
```
rnathuji, line 40: Should you make this an assertion?

tests/s3_bucket/test_github_content_repos.py
```diff
@@ -0,0 +1,52 @@
...
+
+            else:
+                resp_content = requests.get(
```
rnathuji, line 43: I think the previous request already queried the content. Can you access as just `module_resp.content` here?

omehes: @rnathuji This I already tried and I always got this error: AttributeError: 'HTTPResponse' object has no attribute 'content'. Had to use requests.get()...

omehes:
@rnathuji made the recommended changes

tests/s3_bucket/test_github_content_repos.py
rnathuji: Oh, my mistake, it needs to be `module_resp.read()`. That should give you the contents of the CNXML without doing an additional request.

omehes:
@rnathuji done :)

rnathuji (approved):

PR #729, opened 2021-03-26
amendments to the fixture for the github repo tests

omehes:
git_content_repos.py iterates through openstax repos and selects all starting with osbooks-


fixtures/git_content_repos.py
```diff
@@ -1,10 +1,28 @@
...
 @pytest.fixture
-def git_content_repos():
-    """Returns the list of all github content repos
+def git_content_repos(headers_data):
+
+    """Returns the list of all github content repos (starting with osbooks-)
     """
-    repo = ["osbooks-university-physics", "osbooks-college-algebra-bundle"]
+
+    repos_url = "https://api.github.com/orgs/openstax/repos?per_page=100&type=private"
```
rnathuji, line 15: It looks like there are currently 77 private repos. The problem I see with using `per_page=100` versus using paging is if we go over 100 (and I believe 100 is the max one can request using `per_page`), the fixture could start missing repos without any obvious signal that this is happening. I'd suggest adding the paging support now so the fixture works with larger numbers of repos. However, if we leave the implementation this way, perhaps we should at least throw an exception if 100 results are returned as that will tell us the test code likely needs to be updated to support more results?

omehes: yeah. I was thinking the same and wanted your opinion on this. I am also worried about paging. If I add page=4 (where the 2 osbooks- collections can be found) and in the near future more repos will be added, will not some of them fall to page=5? Or maybe at per_page=100 it won't happen that soon...?

rnathuji: It's hard to predict when we might cross 100 since the repos are org wide. I know we'll be adding more for the content repos which will all be private, but I don't know for sure that would put us over 100. 

However, you shouldn't need to use page numbers directly. Something like the following could do the trick as far as iterating all pages (note I'm using `per_page=25` so there will be multiple pages for demonstration / testing):

```python
url = 'https://api.github.com/orgs/openstax/repos?per_page=25&type=private'
resp = requests.get(url, headers=headers)
print(f"Got {len(resp.json())} items")
while resp.links.get('next'):
    next_url = resp.links['next']['url']
    print(f"Getting next page at {next_url}")
    resp = requests.get(next_url, headers=headers)
    print(f"Got {len(resp.json())} items")
```

The next page link is provided in the response header when available, so you can just grab it and iterate that way.

fixtures/git_content_repos.py
```diff
@@ -1,10 +1,36 @@
...
-    return repo
+    url = "https://api.github.com/orgs/openstax/repos?per_page=50&type=private"
+
+    resp = requests.get(url, headers=headers_data)
+
+    repos = []
+
+    while resp.links.get("next"):
+        next_url = resp.links["next"]["url"]
+        resp = requests.get(next_url, headers=headers_data)
```
rnathuji, line 25: The way this loop structure is setup, I think you may actually end up ignoring the first batch of 50 responses (it just so happens the repos you want end up on page 2, so this may not be obvious).

fixtures/git_content_repos.py
```diff
@@ -1,10 +1,36 @@
...
+
+        repos_req = urllib.request.Request(resp.url, headers=headers_data)
+        repos_list = urllib.request.urlopen(repos_req).read()
```
rnathuji, line 28: This is minor, but you end up requesting the same data here as you do in L25 via `requests`. I suspect this may be because you're using `requests` to get the paging links in the header since it parses those automatically for you. However, if you want to reduce this network overhead, you can either:

1. Parse the same headers with urllib manually (e.g. they're available in the response `urllib.request.urlopen(repos_req).headers.get('Link')`). 
2. Remove use of `urllib` altogether in this fixture and drop L27-28 by making L31 `for item in resp.json():`.

fixtures/git_content_repos.py
omehes: @rnathuji Yeees. That is why when I added page=100, it did not give me any repos but with 25 or 50 it gave me 6 osbooks- repos. hm...

omehes:
@rnathuji Apart from the changes you recommended and helped me with, I also added a check for the licenses as per a previous PR from Brenda.

rnathuji (approved):

PR #739, opened 2021-05-24
poet ui smoke test

omehes:
tests poet ui automatically, preferably after new extension deploy
I would like to have devs opinion on the code, especially the login part

### DIFF
--- a/tests/poet/test_poet_ui.py
+++ b/tests/poet/test_poet_ui.py
@@ -1,0 +1,69 @@
+import requests
+
+from urllib.error import HTTPError
+from selenium.common.exceptions import NoSuchElementException
+
+import pytest
+
+from pages.poet.home import Home
+
+import os
+
+"""
+Verifies POET ui
+Latest update on May 21st, 2021
+"""
+
+
+def test_poet_ui(selenium, headers_data):
+    github_user = os.environ.get("GITHUB_USER")
+    github_password = os.environ.get("GITHUB_PASSWORD")
+
+    git_repo_url = os.environ.get("GIT_REPO_URL")
+
+    try:
+        requests.get(git_repo_url, headers=headers_data)
+
+    except HTTPError as h_e:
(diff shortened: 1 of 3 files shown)


tests/poet/test_poet_ui.py
```diff
@@ -0,0 +1,69 @@
...
+    else:
+        # launches gitpod for openstax/poet test repo and logs in
+        home = Home(selenium, git_repo_url).open()
+
+        home.click_github_login_button()
+
+        selenium.switch_to.window(selenium.window_handles[1])
+
+        home.fill_user_field(github_user)
+        home.fill_passwd_field(github_password)
```
rnathuji, line 40: Is the idea / intent that these tests would get run with a dedicated test GH user with limited perms or an individual's real credentials (the latter would only work here if the user doesn't have MFA enabled on their account I think)?

rnathuji (commented):
I took a look at the auth aspect per the PR comment, but I have a general question on the intended scope of this test suite in `cnx-automation`  versus adding to things like the cypress tests in the POET repo itself. Specifically, is there coverage we want to achieve here using selenium that we'd be unable to with the testing frameworks used in the repo (e.g. is the goal to catch gitpod-specific problems somehow, or ...)?

tests/poet/test_poet_ui.py
omehes: What I was thinking is to have an auto smoke test for poet everytime we release a new extension to registry. Basically, new extension is in registry, this set of tests would automatically kick off and test the poet and make sure that it runs. Something similar, what we had in cnx-automation for cnx webview. Just an idea and also for me to practice writing auto-tests :) But it would be good to have it.

omehes: Imagine it as a ui smoke test of poet extension

rnathuji: 👍

rnathuji (approved):

PR #741, opened 2021-06-01
generate requirements.txt from requirements.in

m1yag1:


m1yag1:
I updated each of dependencies that was included in the requirements.in file. This caused a few of the dependencies to be downgraded b/c of the way pip-tools resolves working versions.

rnathuji (approved):

PR #743, opened 2021-06-30
changing/adding assertions for git content repo test

omehes:


rnathuji (approved):
Purely as FYI, if you're curious about how the name spacing issues can be abstracted by parsing the data as XML and asserting on expected content that way, you can look at one of our scripts [here](https://github.com/openstax/content-synchronizer/blob/987db036c43947df3d8e29b6c6d7cebde40770b5/resource-synchronizer/update_metadata.py#L95) which uses [these definitions](https://github.com/openstax/content-synchronizer/blob/main/resource-synchronizer/namespaces.py).

PR #750, opened 2021-08-12
QA search tool to look through all index files in all content repos

omehes:
just looking for an opinion/any issues with it

### DIFF
--- a/tests/s3_bucket/test_search_github_content_repos.py
+++ b/tests/s3_bucket/test_search_github_content_repos.py
@@ -1,0 +1,64 @@
+import requests
+
+import urllib.parse
+from urllib.error import HTTPError
+
+import pytest
+
+"""
+Searches for an input string in index.cnxml of every collection module of every github content repo.
+Latest update on April 19th, 2021
+"""
+
+
+def test_search_github_content_repos(git_content_repos, headers_data):
+
+    tosearch = input("\nEnter string: ")
+
+    for repo in git_content_repos:
+
+        print("\nNow searching modules of: ", repo)
+
+        modules_dir = f"https://api.github.com/repos/openstax/{repo}/contents/modules/"
+
+        try:
+
+            modules_list = requests.get(modules_dir, headers=headers_data)
+
+        except HTTPError as h_e:
+            # Return code 404, 501, ... for incorrect modules url
+            pytest.fail(f"HTTP Error {h_e.code}: incorrect modules url {modules_dir}")
+
+        else:
+
+            for item in modules_list.json():
+
+                if item["type"] != "dir":
+
+                    # Ignore anything that may not be a directory
+                    continue
+
+                rel_path = urllib.parse.quote(item["path"])
+                modules_url = (
+                    f"https://api.github.com/repos/openstax/{repo}/contents/{rel_path}/index.cnxml"
+                )
+
+                try:
+                    module_resp = requests.get(modules_url, headers=headers_data)
+
+                except HTTPError as h_e:
+                    # Return code 404, 501, ... for incorrect/missing index.cnxml
+                    pytest.fail(
+                        f"HTTP Error {h_e.code}: incorrect/missing index.cnxml {modules_url}"
+                    )
+
+                else:
+                    xhtml_data = module_resp.text
+                    found_data = xhtml_data.find(tosearch)
+
+                    if found_data >= 1:
+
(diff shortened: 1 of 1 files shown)


tests/s3_bucket/test_search_github_content_repos.py
```diff
@@ -0,0 +1,75 @@
...
+
+        else:
+
+            print(f'\nSearching for "{string_to_search}" in modules of: ', repo)
+
+            modules_dir = f"https://api.github.com/repos/openstax/{repo}/contents/modules/"
+
+            try:
+
+                modules_list = requests.get(modules_dir, headers=headers_data)
```
rnathuji, line 36: Just a thought, but since you're hitting the github API anyways, would [this endpoint](https://docs.github.com/en/rest/reference/search) be simpler / more efficient? Something like a `GET` to `f"https://api.github.com/search/code?q={string_to_search}+repo:openstax/{repo}"`. There are [additional qualifiers](https://docs.github.com/en/github/searching-for-information-on-github/searching-on-github/searching-code) you might find handy too.

It looks like it does have [some restrictions](https://docs.github.com/en/rest/reference/search#considerations-for-code-search), though.

omehes: @rnathuji this restriction is worrying, as some index files could become larger than 384 KB, no? 
`Only files smaller than 384 KB are searchable.`

rnathuji: Yeah, it's possible. FWIW the largest CNXML in `university-physics` seems to be 252KB ([m58280](https://github.com/openstax/osbooks-university-physics/blob/main/modules/m58280/index.cnxml)). But if you need this search to never give a false negative then you're right about not making any assumptions around max size.

omehes: I checked some index files in college algebra bundle too, found one with 273 kb. We might not have larger files than 384 kb yet but who knows with new books... If file is larger than the max, do you know what would happen (throws an error or just skip to the next one)?

rnathuji: The docs are unclear to me, but I'm guessing be the entire file gets skipped for indexing and it would never show up in results.

omehes: I checked some repos, fizyka-bundle, I got some large(r) files:

```
gitpod /workspace/osbooks-fizyka-bundle/modules $ git ls-tree -r -t -l --full-name HEAD | sort -n -k 4 | tail -n 4
100644 blob 0561aa2f474784d649072ab0070729d7e4dc9ec9  289610    modules/m65744/index.cnxml
100644 blob dd53d60eb5baa089b8f5581f3da65ed85e70842e  293906    modules/m68252/index.cnxml
100644 blob 4ca9641de9582b80d467906186e93a4e79fedbda  328311    modules/m68247/index.cnxml
100644 blob e777a2b31805ab27662816b248e9d3ef5755cdc7  880216    modules/m65745/index.cnxml
```

rnathuji: Yep, and it looks like if you search for `fs-id1167131249217` in that repo, `m65745` doesn't come up. So that confirms our suspicion the file just gets skipped. I think my idea may be a non-starter 😂 .

omehes: yeah, that is not good... what do you think then?

rnathuji: I think sticking with your original approach is probably fine for now so there's no confusion if something doesn't show up due to large file size. 

It occurred to me in case it's useful down the road: This may (or may not) relate to [sifter](https://github.com/openstax/sifter) in as much as @philschatz may have plans to do a git storage version of that 🤷‍♂️ .

omehes: @rnathuji would you approve it (for now). I checked Phil's creation. Looks great but still under "construction" :)

rnathuji (approved):

tests/s3_bucket/test_search_github_content_repos.py
rnathuji: Oh, d'oh! I thought I already had, sorry. Approved now 😅 .

omehes: no worries. thanks :)

PR #755, opened 2021-09-09
amending qa redirect test to check for 200 and non-200 status codes for each page url variation

omehes:
these are amendments to my redirect test in cnx-automation repo (test_rex_redirects.py) as per discussions from few weeks back. As per Phil's comments I removed checking page slugs and added checking for response.status_code. Anything different than 200 is then written out to a csv file which then can be checked by Neil and others to investigate while the book page fails. I assign this to @philschatz as per his explanation/advise and/or @rnathuji

### DIFF
--- a/tests/webview/integration/test_rex_redirects5.py
+++ b/tests/webview/integration/test_rex_redirects5.py
@@ -1,0 +1,50 @@
+import backoff
+import requests
+
+from tests import markers
+
+import csv
+import pytest
+
+
+@backoff.on_exception(backoff.expo, requests.exceptions.ConnectionError)
+def get_url(url):
+    return requests.get(url)
+
+
+@markers.slow
+@markers.nondestructive
+def test_calculus2_uri_redirect_to_rex(webview_base_url, rex_base_url, calculus_vol_2_uri):
+    # GIVEN a webview_base_url, rex_base_url and a calculus_vol_2_uri
+
+    # WHEN we go to a page based on the webview_base_url and uri
+    cnx_page_slug = calculus_vol_2_uri.split("/")[-1]
+    cnx_url = f"{webview_base_url}{calculus_vol_2_uri}"
+    response = get_url(cnx_url)
+
+    status_codes = []
+    error_urls = []
+
+    # THEN we are redirected to rex
+    if response.status_code != 200:
+
+        status_codes.append(str(response.status_code))
+        error_urls.append(cnx_url)
+
+        dict_urls = {"PAGE URL": error_urls[i] for i in range(len(status_codes))}
+        dict_errors = {"ERROR": status_codes[i] for i in range(len(status_codes))}
+        dict_urls_errors = {**dict_urls, **dict_errors}
+
+        with open("url_error_list.csv", "a") as urlerr:
+            writer = csv.DictWriter(urlerr, dict_urls_errors.keys())
+
+            if urlerr.tell() == 0:
+                writer.writeheader()
+
+            writer.writerow(dict_urls_errors)
+
+        pytest.fail(f"{response.status_code} in {cnx_page_slug}")
+
+    else:
+        assert 200 == response.status_code
+        assert response.url.startswith(rex_base_url)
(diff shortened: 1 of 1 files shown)


omehes:
Note, that ones these changes are approved, I will copy this code into the original test_rex_redirects.py file for each book entry. This new file here is only to create and discuss my changes.

tests/webview/integration/test_rex_redirects5.py
```diff
@@ -0,0 +1,50 @@
...
+
+            if urlerr.tell() == 0:
+                writer.writeheader()
+
+            writer.writerow(dict_urls_errors)
+
+        pytest.fail(f"{response.status_code} in {cnx_page_slug}")
+
+    else:
+        assert 200 == response.status_code
```
rnathuji, line 49: I think this `assert` will always pass and can be removed

tests/webview/integration/test_rex_redirects5.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # THEN we are redirected to rex
+    if response.status_code != 200:
+
+        status_codes.append(str(response.status_code))
+        error_urls.append(cnx_url)
+
+        dict_urls = {"PAGE URL": error_urls[i] for i in range(len(status_codes))}
+        dict_errors = {"ERROR": status_codes[i] for i in range(len(status_codes))}
+        dict_urls_errors = {**dict_urls, **dict_errors}
```
rnathuji, line 36: I may just be looking at this wrong, but won't `len(status_codes)` and `len(error_urls)` always be one? That is, could this be simplified to the following?:

```python
dict_url_errors = {
    "PAGE URL": response.status_code,
    "ERROR": cnx_url
}
```

tests/webview/integration/test_rex_redirects5.py
```diff
@@ -0,0 +1,50 @@
...
+
+        with open("url_error_list.csv", "a") as urlerr:
+            writer = csv.DictWriter(urlerr, dict_urls_errors.keys())
+
+            if urlerr.tell() == 0:
+                writer.writeheader()
+
+            writer.writerow(dict_urls_errors)
+
+        pytest.fail(f"{response.status_code} in {cnx_page_slug}")
```
rnathuji, line 46: As part of merging into `test_rex_redirects.py`, you can probably keep things DRY by pulling this out to a helper function like you did with `get_url`.

tests/webview/integration/test_rex_redirects5.py
omehes: yeah, I was thinking the same just wanted to see other opinions on it

tests/webview/integration/test_rex_redirects5.py
omehes: @rnathuji I made the recommended changes but this one (helper function) I am not sure. Looking at it but don't see how...?

rnathuji: 🤔 - Maybe you could do something like:

```python
def log_failure_to_csv(status_code, cnx_url):
    dict_urls_errors = {"ERROR": status_code, "PAGE URL": cnx_url}

    with open("url_error_list.csv", "a") as urlerr:
        writer = csv.DictWriter(urlerr, dict_urls_errors.keys())
        if urlerr.tell() == 0:
            writer.writeheader()
        writer.writerow(dict_urls_errors)
```

...and then invoke as:

```python
    # THEN we are redirected to rex
    if response.status_code != 200:
        log_failure_to_csv(response.status_code, cnx_url)
        pytest.fail(f"{response.status_code} in {cnx_page_slug}")
    else:
        assert response.url.startswith(rex_base_url)
```

Or at least I think that will reduce some repetition when you move this test structure over to `test_rex_redirects.py`, but you can see what you think when doing that.

rnathuji (approved):

tests/webview/integration/test_rex_redirects5.py
omehes: hm... I was trying something similar but I was getting unknown parameters/variables... thanks for the tip and review :)

omehes: @rnathuji thanks for the suggestion. Now I see where I went wrong when I tried the helper function. here are the changes. Also, I noticed that certain 404 errors will always occur because some urls lead to main chapters but they are not real pages. I am looking at the `ce-scripts/.venv_scripts/lib/python3.8/site-packages/rex_redirects.py` where all the url variants are created. If we could eliminate all fake pages (I think they start with #) in there, it would reduce many 404 failures and make the report more accurate.

omehes:
@rnathuji ah, you already approved :)

PR #758, opened 2021-09-17
amending and adding more checks for content repos

omehes:
added tests to:
verify presence of meta-inf folder
verify presence of media folder

made some changes to verification code

renamed: test_github_content_repos.py -> test_github_content_modules.py

### DIFF
--- a/tests/s3_bucket/test_github_content_media.py
+++ b/tests/s3_bucket/test_github_content_media.py
@@ -1,0 +1,39 @@
+import requests
+
+"""
+Verifies that media folder in every github content repo exists and is not empty.
+Latest update on September 16th, 2021
+"""
+
+
+def test_github_content_media(git_content_repos, headers_data):
+
+    html_url_count = []
+
+    for repo in git_content_repos:
+
+        print("\nNow verifying: ", repo)
+
+        media_dir = f"https://api.github.com/repos/openstax/{repo}/contents/media/"
+
+        media_list = requests.get(media_dir, headers=headers_data)
+
+        if media_list.status_code != 200:
+
+            # Return code 404, 501, ... for incorrect media url
+            print(f">>>>> FAILED {media_list.status_code}: no media folder in {repo}")
+
+        else:
+
+            for item in media_list.json():
+
+                if item["type"] != "file":
+
+                    # Ignore anything that may not be a file
+                    continue
+
+                else:
+
+                    html_url_count.append(item["html_url"])
+
+        assert sum(".jpg" in s for s in html_url_count) > 0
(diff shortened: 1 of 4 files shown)


tests/s3_bucket/test_github_content_media.py
```diff
@@ -0,0 +1,39 @@
...
+                if item["type"] != "file":
+
+                    # Ignore anything that may not be a file
+                    continue
+
+                else:
+
+                    html_url_count.append(item["html_url"])
+
+        assert sum(".jpg" in s for s in html_url_count) > 0
```
rnathuji, line 39: Question / thought: I believe this assumes a valid media directory will have a `*.jpg` file in it. However: 1) There might be cases for new books early in development where there are some media files but they don't happen to be of that extension; 2) I think there early on it's valid for the media directory to only have a `.gitkeep` file (maybe worth asking / confirming with the team). In both cases this assertion may lead to false positive failures.

tests/s3_bucket/test_github_content_meta_inf.py
```diff
@@ -0,0 +1,60 @@
...
+                        pytest.fail(
+                            f"HTTP Error {h_e.code}: incorrect/missing books.xml {meta_inf_url}"
+                        )
+
+                    else:
+
+                        resp_content = meta_inf_resp.text
+
+                        # Verifies books.xml files for presence of content
+                        assert resp_content.count("<book slug") > 0
```
rnathuji, line 60: This assertion might be problematic as it combines the element name and an attribute, but the attribute could come later. For example, both of the following would be valid:

```xml
<book slug="book1" collection-id="col12345" href="../collections/book1.collection.xml" />
```

```xml
<book collection-id="col12345" slug="book1" href="../collections/book1.collection.xml" />
```

If you're trying to avoid parsing the file as XML, perhaps just doing `.count("<book")` would be sufficient 🤷‍♂️ ?

tests/s3_bucket/test_github_content_media.py
omehes: Thanks. You are right. Maybe I should check for any kind of file in the folder or just that the folder exists... but I would like to know if there is something in it as well in case there were changes and the media files were deleted or not copied back or left out completely... Yeah, the development phase might be a problematic phase...

tests/s3_bucket/test_github_content_meta_inf.py
omehes: You are right here too. I will change it to only check for <book

omehes:
@rnathuji amended the tests and added some checks for (empty) media folder (.gitkeep is added automatically when using the repo template).

rnathuji (approved):

PR #769, opened 2022-02-22
corgi e2e tests written in playwright

omehes:
for QA written in playwright: tests to verify corgi jobs and artifacts (pdf/preview)

### DIFF
--- a/fixtures/corgi_playwright.py
+++ b/fixtures/corgi_playwright.py
@@ -1,0 +1,39 @@
+from playwright.sync_api import sync_playwright
+import pytest
+
+
+@pytest.fixture
+def chrome_browser():
+
+    playwright_sync = sync_playwright().start()
+    chromebrowser = playwright_sync.chromium.launch(headless=False, slow_mo=1000)
+
+    return chromebrowser
+
+
+@pytest.fixture
+def chrome_page(chrome_browser, corgi_user, corgi_password):
+    context = chrome_browser.new_context(
+        http_credentials={"username": corgi_user, "password": corgi_password}
+    )
+    page = context.new_page()
+
+    return page
+
+
+@pytest.fixture
+def corgi_user(request):
+    """Return value of the corgi username"""
+    config = request.config
+    corgi_user = config.getoption("corgi_user") or config.getini("corgi_user")
+    if corgi_user is not None:
+        return corgi_user
+
+
+@pytest.fixture
+def corgi_password(request):
+    """Return value of the corgi password"""
+    config = request.config
+    corgi_password = config.getoption("corgi_password") or config.getini("corgi_password")
+    if corgi_password is not None:
+        return corgi_password
--- a/pages/corgi/home.py
+++ b/pages/corgi/home.py
@@ -1,0 +1,139 @@
+class Home:
+    def __init__(self, page):
+        self.page = page
+
+    @property
+    def create_new_job_button_locator(self):
+        return self.page.wait_for_selector("text=create a new job")
+
+    def click_create_new_job_button(self):
+        self.create_new_job_button_locator.click()
+
+    @property
+    def pdf_radio_button_locator(self):
+        return self.page.wait_for_selector("div.v-radio.pdf-radio-button")
+
+    def click_pdf_radio_button(self):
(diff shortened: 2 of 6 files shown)


omehes:
The fails are not related to my code

fixtures/corgi_playwright.py
```diff
@@ -0,0 +1,39 @@
...
+    context = chrome_browser.new_context(
+        http_credentials={"username": corgi_user, "password": corgi_password}
+    )
+    page = context.new_page()
+
+    return page
+
+
+@pytest.fixture
+def corgi_user(request):
```
sparksam, line 25: We can have a method 
```
def attribute_from_request(request, attribute):
    config_option = request.config.getoption(attribute)
    return config_option if config_option else request.config.getini(attribute)
  
@pytest.fixture
def corgi_user(request):
     return attribute_from_request(request, "corgi_user")
     
@pytest.fixture
def corgi_password(request):
     return attribute_from_request(request, "corgi_password")

File: fixtures/corgi_playwright.py [hunk 2]
```diff
@@ -0,0 +1,39 @@
+from playwright.sync_api import sync_playwright
```
m1yag1, line 1: These tests should not be created within the cnx-automation repository. If we're writing tests they should live within the same repository as the source code [here](https://github.com/openstax/output-producer-service/blob/main/backend/app/tests/)

m1yag1 (dismissed):
This needs to be moved to the CORGI repository

fixtures/corgi_playwright.py
omehes: as per our discussion, I will use these tests as my test-tools in cnx-automation but once all the UI tests were rewritten in playwright, I will look into moving these too into output-producer-service

pages/corgi/home.py
```diff
@@ -0,0 +1,139 @@
...
+
+    @property
+    def create_button_locator(self):
+        return self.page.locator("button.create-button-start-job")
+
+    def click_create_button(self):
+        self.create_button_locator.locator("text=Create").click()
+        self.create_button_locator.locator("text=Create").click()
+
+    def click_create_button_pdf(self):
```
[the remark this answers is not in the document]
omehes: @sparksam yeah, they are the same but for some reason, click on Create button has to be done twice when running git and git-preview jobs. I cannot figure out why???

pages/corgi/home.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def click_content_server_dropdown(self, value):
+        self.content_server_dropdown_locator.locator(f"text={value}").click()
+
+    @property
+    def create_button_locator(self):
+        return self.page.locator("button.create-button-start-job")
+
+    def click_create_button(self):
+        self.create_button_locator.locator("text=Create").click()
```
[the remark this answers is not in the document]
omehes: ok. I'll use this. thanks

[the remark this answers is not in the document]
omehes: @sparksam This is what I had to do for git/git-preview jobs for a succesful Create button click

```    def click_create_button(self):
        button = self.create_button_locator
        if button:
            button.click()
            button.click()
```
Don't know why this happens. Cannot figure this out...

fixtures/corgi_playwright.py
omehes: I will look into this but for now I leave it as is, if you don't mind

fixtures/corgi_playwright.py
m1yag1: Ok, Otto and I discussed this a bit further. The overarching plan atm is to use cnx-automation to learn playwright and develop these tests that are useful for Otto. At the same time, Otto will work to replacing the selenium tests in CORGI with Playwright 1 for 1 until all the tests are done. At that point, we'll try to work more of the tests that Otto has in cnx-automation into CORGI so that we can test the entire pipeline after a deployment.

fixtures/corgi_playwright.py
sparksam: Not a problem.

omehes:
closing as these will be moved to output-prod-service

PR #777, opened 2022-04-14
adding code to run corgi git jobs easily (+ international collections) 

omehes:
this is to run my smoke test on corgi after releases


omehes:
This is outdated. I have more changes to come. It's my test tool to create corgi jobs when checking corgi release on staging and prod

PR #781, opened 2022-05-06
amending assertions in test_github_content_modules_1.py and removing test repos from list

omehes:


tests/s3_bucket/test_github_content_modules_1.py
```diff
@@ -55,9 +59,45 @@ def test_github_content_modules_1(git_content_repos_1, headers_data):
...
-                    )
+                    soup = BeautifulSoup(resp_content, "xml")
+                    document = soup.find_all("document")
+                    content = soup.find_all("content")
+
+                    for doc in document:
+
+                        # Verifies index.cnxml for presence of md:title and md:uuid metadata
+                        try:
+                            md_title = re.findall("<md:title>(.*)</md:title>", str(doc))
```
TylerZeroMaster, line 70: Why use regex when you already have `BeautifulSoup`? This could be `assert doc.find("md:title")`. Also, the standard `re.findall` will return an empty array when it does not find anything, not `None`. So it will never be `None`.

TylerZeroMaster (changes requested):

tests/s3_bucket/test_github_content_modules_1.py
omehes: you are right... on it

omehes:
@TylerZeroMaster Hi. Here are the changes you recommended. Ready for your review if you have time

tests/s3_bucket/test_github_content_modules_1.py
```diff
@@ -55,9 +58,47 @@ def test_github_content_modules_1(git_content_repos_1, headers_data):
...
+                    content = soup.find_all("content")
+
+                    for doc in document:
+
+                        md_title = doc.find_all("md:title")
+                        md_content_id = doc.find_all("md:content-id")
+
+                        # Verifies index.cnxml for presence of md:title and md:uuid metadata
+                        try:
+                            assert md_title is not None
```
TylerZeroMaster, line 72: Sorry to say, but this still has a problem similar to before: `doc.find_all("...")` will always return a list of zero or more elements. If you ran `doc.find_all("element_that_does_not_exist")`, you would get `[]`.

If you ran this on a doc that was missing a `md:title` element, the assertion would not raise an exception because you would get an empty list and an empty list `is not None`. You could check if the length of `md_title` is greater than 0 (or maybe equal to 1 if you want to make sure that it only has one title). Alternatively, you could use `doc.find` instead which returns either the first matching element or `None`.

TylerZeroMaster (approved):

PR #785, opened 2022-05-09
amending test_github_content_modules_5.py - adding bs4 and changing assertions

omehes:


omehes:
@TylerZeroMaster last one. THANKS!

PR #786, opened 2022-05-09
amending test_github_content_meta_inf.py to use bs4 and excluding test repos in fixture

omehes:


omehes:
@TylerZeroMaster I added you to this PR as it is related to the ones you reviewed but if you busy you can leave it to others :) one more similar PR will be coming, as I realised that I have the same issue as you previously pointed out - find_all is not None (when empty)...

TylerZeroMaster (approved):

PR #787, opened 2022-05-10
amending assertions in test_github_content_collections.py

omehes:


omehes:
@TylerZeroMaster This is the last test I wanted to amend and change assertions (find_all to find, etc)

tests/s3_bucket/test_github_content_collections.py
```diff
@@ -65,48 +65,52 @@ def test_github_content_collections(git_content_repos, headers_data, abl_books_u
...
+                    col_collection = soup.find_all("collection")
 
-                    md_title = soup.find_all("md:title")
-                    md_license = soup.find_all("md:license")
-                    content = soup.find_all("col:content")
+                    for col in col_collection:
 
-                    try:
+                        try:
+                            assert col.find("md:title").text is not None
```
TylerZeroMaster, line 73: This might be a problem here and on line 86: col.find(...).text will return an empty string, not None. 

If you want to check that the element has content, it is better to test the length (I usually throw in a `.strip()` too in case someone has included a space inside an element.
```python
>>> col = BeautifulSoup('<root><a></a></root>')
>>> col.find('a').text
''
>>> # this one would never raise an assertion error
>>> assert col.find('a').text is not None
>>>
>>> assert len(col.find('a').text.strip()) > 0
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AssertionError
>>>
>>> col = BeautifulSoup('<root><a>        </a></root>')
>>> assert len(col.find('a').text.strip()) > 0
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AssertionError
```

TylerZeroMaster (changes requested):

tests/s3_bucket/test_github_content_collections.py
omehes: ok. first I tried len than I felt that None is better... :)

omehes:
@TylerZeroMaster I changed the assertions. It is definately better the way you proposed. It gives me assertion error for 'col:content' in world-history (and some others) as it is empty and when checked, they are really empty...

omehes:
@TylerZeroMaster I made the changes as per your review :)

TylerZeroMaster (approved):

PR #788, opened 2022-05-26
changing private to public in github api in all github test fixtures + removing non-existent fixture

omehes:


fixtures/git_content_repos.py
```diff
@@ -18,7 +18,7 @@ def git_content_repos(headers_data):
         "osbooks-playground",
     ]
 
-    next_url = "https://api.github.com/orgs/openstax/repos?per_page=50&type=private"
+    next_url = "https://api.github.com/orgs/openstax/repos?per_page=50&type=public"
```
TylerZeroMaster, line 21: Only the books on the ABL are public. If you want to get a list of all the books, you will need to search public and private repositories.

The `type` param is optional. Could you use `https://api.github.com/orgs/openstax/repos?per_page=50` (without type), so you can get public and private repositories at the same time?

TylerZeroMaster (approved):

PR #790, opened 2022-06-06
adding test to compare slugs/styles of content repos against abl json

omehes:


tests/s3_bucket/test_github_content_compare_slugs_styles.py
```diff
@@ -0,0 +1,107 @@
...
+
+                    except HTTPError as h_e:
+                        # Return code 404, 501, ... for incorrect/missing books.xml
+                        pytest.fail(
+                            f"HTTP Error {h_e.code}: incorrect/missing books.xml {meta_inf_url}"
+                        )
+
+                    else:
+
+                        resp_content = meta_inf_resp.text
```
TylerZeroMaster, line 62: This request [returns a json document](https://docs.github.com/en/rest/repos/contents). Within that json document is a property called `contents` that is base64 encoded. You can decode the value of this property and use the xml document that way:

```python
from base64 import b64decode
json_resp = meta_inf_resp.json()
resp_content = b64decode(json_resp["content"])
```

tests/s3_bucket/test_github_content_compare_slugs_styles.py
```diff
@@ -0,0 +1,107 @@
...
+
+def test_github_content_compare_slugs_styles(
+    git_content_repos, headers_data, abl_books_slugs_styles
+):
+
+    for repo in git_content_repos:
+
+        print("\nNow verifying: ", repo)
+
+        meta_inf_dir = f"https://api.github.com/repos/openstax/{repo}/contents/META-INF/"
```
TylerZeroMaster, line 25: Keep in mind that the version of content on `main` may not match the version that has been added to the ABL. If that is important, you can include the `commit_sha` from the ABL in `abl_books_slugs_styles` and then use that with the `ref` query argument like 
```python
meta_inf_dir = f"https://api.github.com/repos/openstax/{repo}/contents/META-INF?ref={commit_sha}"
```

The way you are running this test right now would check if the `META-INF/books.xml` on `main` contains information that also exists anywhere in the ABL.

TylerZeroMaster (changes requested):

tests/s3_bucket/test_github_content_compare_slugs_styles.py
omehes: @TylerZeroMaster I tried `json_resp = meta_inf_resp.json()` before but it gives an error (seems like not a json), so I went with .text which works...

tests/s3_bucket/test_github_content_compare_slugs_styles.py
omehes: @TylerZeroMaster hm... I am confused by this... how books.xml and ABL relate to each other? Should they match on main?

tests/s3_bucket/test_github_content_compare_slugs_styles.py
TylerZeroMaster: I don't know what to say. I ran this locally step-by-step and the result was...
```python
>>> meta_inf_resp.text
'{"name":"books.xml","path":"META-INF/books.xml","sha":"71b6b7d577e5b4180e4e8750da622c9f52bcbe88","size":368,"url":"https://api.github.com/repos/openstax/osbooks-chemistry-bundle/contents/META-INF/books.xml?ref=main","html_url":"https://github.com/openstax/osbooks-chemistry-bundle/blob/main/META-INF/books.xml","git_url":"https://api.github.com/repos/openstax/osbooks-chemistry-bundle/git/blobs/71b6b7d577e5b4180e4e8750da622c9f52bcbe88","download_url":"https://raw.githubusercontent.com/openstax/osbooks-chemistry-bundle/main/META-INF/books.xml","type":"file","content":"PGNvbnRhaW5lciB4bWxucz0iaHR0cHM6Ly9vcGVuc3RheC5vcmcvbmFtZXNw\\nYWNlcy9ib29rLWNvbnRhaW5lciIgdmVyc2lvbj0iMSI+CiAgICA8Ym9vayBz\\nbHVnPSJjaGVtaXN0cnktMmUiIGNvbGxlY3Rpb24taWQ9ImNvbDI2MDY5IiBz\\ndHlsZT0iY2hlbWlzdHJ5IiBocmVmPSIuLi9jb2xsZWN0aW9ucy9jaGVtaXN0\\ncnktMmUuY29sbGVjdGlvbi54bWwiIC8+CiAgICA8Ym9vayBzbHVnPSJjaGVt\\naXN0cnktYXRvbXMtZmlyc3QtMmUiIGNvbGxlY3Rpb24taWQ9ImNvbDI2NDg4\\nIiBzdHlsZT0iY2hlbWlzdHJ5IiBocmVmPSIuLi9jb2xsZWN0aW9ucy9jaGVt\\naXN0cnktYXRvbXMtZmlyc3QtMmUuY29sbGVjdGlvbi54bWwiIC8+CjwvY29u\\ndGFpbmVyPgo=\\n","encoding":"base64","_links":{"self":"https://api.github.com/repos/openstax/osbooks-chemistry-bundle/contents/META-INF/books.xml?ref=main","git":"https://api.github.com/repos/openstax/osbooks-chemistry-bundle/git/blobs/71b6b7d577e5b4180e4e8750da622c9f52bcbe88","html":"https://github.com/openstax/osbooks-chemistry-bundle/blob/main/META-INF/books.xml"}}'
```

TylerZeroMaster: It was because of the header `Accept: application/vnd.github.v3.raw` in headers_data fixture. This works as you had it.

tests/s3_bucket/test_github_content_compare_slugs_styles.py
```diff
@@ -0,0 +1,107 @@
...
+
+                    try:
+
+                        meta_inf_resp = requests.get(meta_inf_url, headers=headers_data)
+
+                    except HTTPError as h_e:
+                        # Return code 404, 501, ... for incorrect/missing books.xml
+                        pytest.fail(
+                            f"HTTP Error {h_e.code}: incorrect/missing books.xml {meta_inf_url}"
+                        )
```
TylerZeroMaster, line 58: Don't you need to call `meta_inf_resp.raise_for_status()` for this exception to be raised?

tests/s3_bucket/test_github_content_compare_slugs_styles.py
TylerZeroMaster: I gave this more thought and it might not be a concern. The way it is now, you would only ever test the latest edition that has not yet been added to the ABL; however, I am not sure why you would need to test older editions.

The only other concern is that the contents of META-INF/books.xml on main may contain a style/slug that is not contained in the ABL yet. Granted, it might never be a problem.

tests/s3_bucket/test_github_content_compare_slugs_styles.py
omehes: changes made

TylerZeroMaster (approved):

PR #791, opened 2022-06-07
altering try/except as per review from Tyler

omehes:


omehes:
@TylerZeroMaster I am checking my other tests and altering that try/except which you pointed out in the other PR...

TylerZeroMaster (approved):

PR #792, opened 2022-06-07
altering try/except in github tests as per review

omehes:


omehes:
@TylerZeroMaster the same fix for another test. There will be another PR for all the module tests... and that's it

TylerZeroMaster (approved):

PR #793, opened 2022-06-08
Modus try

omehes:


omehes:
@TylerZeroMaster Another of the try/except changes. This one I had to alter differently... If this is OK, I will create a last PR containing all 4 remaining module tests (same changes in all of them)

TylerZeroMaster (approved):
Do you think it is worthwhile to print `module_resp.status_code` when it falls inside the range?

PR #803, opened 2023-05-19
adding a new content_modules test for testing book repos modules

omehes:
I expanded the index range as there are more books now. Plus, had to update some files...


TylerZeroMaster (commented):
I think the tests in the [circleci config](https://github.com/openstax/cnx-automation/blob/excepts/.circleci/config.yml) can be removed.

TylerZeroMaster (approved):

PR #809, opened 2023-06-26
repo cleanup: removing cnx, archive, legacy tests

omehes:


tests/archive/integration/test_redirects.py
TylerZeroMaster, line 1: All these redirects should still work for cnx.org, but they may work differently. For example, https://cnx.org/contents/36004586-651c-4ded-af87-203aca22d946 goes to internet archive.

TylerZeroMaster (approved):
I am not very familiar with the old tests, but this looks good.

